PR #22, opened 2017-09-17
Simplify spec by moving to capabilities, proofs, and authn and authz

msporny:
Attempt to simplify the spec by moving to:

 * Authentication and Authorization (vs. Ownership and Control)
 * Proofs (vs. just Keys and Signatures)
 * Decomposing Service Discovery
 * Capabilities-based Security Model

You can view the rendered changes here:

https://msporny.github.io/did-spec/


ChristopherA:
@msporny Do you have a URL so we can look at your branch directly?

msporny:
@ChristopherA Yes, here it is:

https://msporny.github.io/did-spec/

I also added it to the PR description just now.

index.html
```diff
@@ -94,17 +92,19 @@
...
 under the control of the identity owner and not dependent on a
 centralized registry, identity provider, or certificate authority. DIDs
-resolve to DDOs (DID descriptor objects)—simple JSON documents that
-contain all the metadata needed to prove ownership and control of a DID.
-Specifically, a DDO contains a set of key descriptions, which are
-machine-readable descriptions of the identity owner’s public keys, and a
-set of service endpoints, which are resource pointers necessary to
-initiate trusted interactions with the identity owner. Each DID uses a
-specific DID method, defined in a separate DID method specification, to
+resolve to DID Objects — simple JSON documents that
```
dlongley: I'm a +1 to changing `DID Objects` to `DID Descriptions`.

index.html
```diff
@@ -94,17 +92,19 @@
...
+contain all the metadata needed to interact with the DID.
```
dlongley: "Interact with the DID" ... how does one "interact" with the DID? Perhaps this should be "with the entity that the DID identifies." 

We often talk about "authenticating as a DID" or "interacting with the DID" as shorthand for "the entity the DID identifies". I know this is verbose, but we shouldn't let the shorthand we use to discuss DIDs leak into this spec because it's inaccurate and confusing to those outside the community that discusses this on a regular basis.

index.html
```diff
@@ -94,17 +92,19 @@
...
+Specifically, a DID Object contains at least three things. The first is
+a set of mechanisms that may be used to authenticate as
```
dlongley: I think we should be careful here to indicate that the DID Description specifies the set of *acceptable* mechanisms or the verification material (not the mechanisms themselves or the authentication secrets). Also, again, one does not authenticate as a particular DID, they authenticate as the entity the DID identifies. So suggested language:

"The first is the set of acceptable forms of proof that may used to authenticate as the entity identified by a DID."

I'd be happy to see something more concise than that, but that's my best stab at the moment.

index.html
```diff
@@ -94,17 +92,19 @@
...
+as a particular DID (e.g. public keys, pseudonymous biometric templates, etc.).
+The second is a set of authorization information that outlines which entities
```
dlongley: "outlines" => "specifies".

index.html
```diff
@@ -94,17 +92,19 @@
...
+may modify the DID Object. The third is a set of service endpoints, which may
+be used to initiate trusted interactions with the identity owner. Each DID
```
dlongley: "with the identity owner" => "with the entity identified by the DID".

index.html
```diff
@@ -174,12 +174,9 @@
 Each identity owner can be identified on a ledger with a <a href=
```
dlongley: "Each identity owner can be identified on a ledger" => "Each entity may authenticate with a"

The entities are identified by DIDs. They may authenticate via keys (or other mechanisms).

dlongley: We should square this language with the more general use of "proofs" perhaps by saying "the simplest authentication mechanism is through the use of a public/private key pair".

index.html
```diff
@@ -356,102 +351,318 @@
...
+        </p>
+      </td>
+      <td colspan="1" rowspan="1">
+        <p>
+          Self-Sovereignty
+        </p>
+      </td>
+      <td colspan="1" rowspan="1">
+        <p>
+          DID architecture should give identity owners the power to
```
dlongley: "identity owners" => "entities"?

index.html
```diff
@@ -356,102 +351,318 @@
...
+        </p>
+      </td>
+      <td colspan="1" rowspan="1">
+        <p>
+          Privacy
+        </p>
+      </td>
+      <td colspan="1" rowspan="1">
+        <p>
+          DID architecture should enable identity owners to control the
```
dlongley: "identity owners" => "entities"?

index.html
```diff
@@ -356,102 +351,318 @@
...
+        </p>
+      </td>
+      <td colspan="1" rowspan="1">
+        <p>
+          Security
+        </p>
+      </td>
+      <td colspan="1" rowspan="1">
+        <p>
+          DID architecture should provide sufficient security for relying
```
dlongley: The architecture should "provide" this security or "enable" it?

index.html
```diff
@@ -356,102 +351,318 @@
...
+      <td colspan="1" rowspan="1">
+        <p>
+          Proof-based
+        </p>
+      </td>
+      <td colspan="1" rowspan="1">
+        <p>
+          DID architecture should enable an identity owner to provide
+          cryptographic proof of ownership and proof of access control
+          rights.
```
dlongley: Uses old language of "proof of ownership" and "proof of access control". Possibly change to:

"DID architecture should enable an entity to authenticate and obtain authorization to modify DID Descriptions."

Though, are there other capabilities beyond modifying the DID Descriptions -- how do we incorporate that language here?

index.html
```diff
@@ -356,102 +351,318 @@
...
-<h2>Example Owner-Managed DDO</h2>
+<h2>Owner-Managed DID Description</h2>
 
 <p>
-Following is an example of a DDO that describes the DID above. This
+Following is an example of a DID Description that describes the DID above. This
 example assumes that the identity owner—the entity that controls the
-private keys for this identity—is authoritative for the DDO. See section
-2.2 for an example of a DDO created by a guardian. For the authoritative
+private keys for this identity—is authoritative for the DID Description. See section
```
dlongley: "identity" => "entity"

index.html
```diff
@@ -356,102 +351,318 @@
...
 </p>
 
 <section>
-<h2>Example Owner-Managed DDO</h2>
+<h2>Owner-Managed DID Description</h2>
 
 <p>
-Following is an example of a DDO that describes the DID above. This
+Following is an example of a DID Description that describes the DID above. This
 example assumes that the identity owner—the entity that controls the
```
dlongley: Old language here that should probably be removed ("identity owner" and "controls") and just talk about entities, authentication, and authorization.

index.html
jandrieu: Also, I think it is a bit imprecise to say that one "authenticates as a particular DID". One does not assert "I am this DID"--as the current language suggests. Rather, we use a DDO to authenticate as an entity authorized to claim the DID. Or something like that. We are not our DIDs. DIDs are merely tools we use to authenticate as appropriate agents in certain circumstances.

msporny:
> I started adding a bunch of +1 to @dlongley's comments for replacing "identity owner" with "entity" and realized a simpler, but more sweeping change to focus on DIDs as providing a means to prove control of "identifiers" rather than "identities" would greatly improve the entire spec.

I agree, but expect this change to be controversial. We should probably discuss in another PR. We may want to raise an issue for this... could you please do that, Joe?

index.html
dlongley: @jandrieu -- I agree. I mentioned the same thing in a couple of comments above. It's seeped into our language when talking about this stuff as shorthand, but it shouldn't be in the spec (and it's arguable that we should even be talking that way about it at all).

jandrieu:
Done, @msporny . As issue #23.

index.html
```diff
@@ -356,102 +348,318 @@
...
-<h2>Example Owner-Managed DDO</h2>
+<h2>Owner-Managed DID Description</h2>
 
 <p>
-Following is an example of a DDO that describes the DID above. This
-example assumes that the identity owner—the entity that controls the
-private keys for this identity—is authoritative for the DDO. See section
-2.2 for an example of a DDO created by a guardian. For the authoritative
+Following is an example of a DID Description that describes the DID above. This
+example assumes that the entity—the entity that controls the
```
kimdhamilton: minor: "entity—the entity" -> "entity"

index.html
```diff
@@ -1110,13 +1074,13 @@
...
 <p>
-A DID MUST be persistent and immutable, i.e., bound to an identity owner
+A DID MUST be persistent and immutable, i.e., bound to an entity
 once and never changed (forever). Ideally a DID would be a completely
 abstract decentralized identifier (like a UUID) that could be bound to
 multiple underlying distributed ledgers or networks over time, thus
 maintaining its persistence independent of any particular ledger or
 network. However registering the same identifier on multiple ledgers or
-networks introduces extremely hard identity ownership and <a href=
+networks introduces extremely hard entityship and <a href=
```
kimdhamilton: is this a find/replace term? replace with "entity ownership"?

index.html
```diff
@@ -528,13 +733,13 @@
 </p>
 
 <p>
-DID descriptor object (DDO). A JSON data structure containing metadata
-describing an identity owner, including the cryptographic key material
-required for the identity owner to prove ownership and control of the
-DID record. A DDO may also contain other <a href=
+DID Description. A JSON data structure containing metadata
+describing a DID, including mechanisms, such as public keys and
+pseudonymous biometrics, that an entity can use to authenticate itself as the
```
devin-fisher: Biometrics on the ledger seems fairly risky.  You can't rotate a biometric if leaked. Crypto algorithms have yet to have as long a life as humans so protecting them with crypto does not have a great track record over a long period of time (a person's life).

ChristopherA:
When we approve these changes in the CG, can we change the spec version to 1.01 or 1.1?

msporny:
> When we approve these changes in the CG, can we change the spec version to 1.01 or 1.1?

No, we can't... a 1.0 release is a W3C REC (if all goes well, that happens in two years time). We could do semantic versioning and back out the version number to 0.x.x, though. We should discuss on the call today. I'll back out the version to 0.7.0

did-v1.jsonld
```diff
@@ -0,0 +1,64 @@
...
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+
+    "CryptographicKey": "sec:Key",
+    "EquihashProof2017": "sec:EquihashProof2017",
+    "GraphSignature2012": "sec:GraphSignature2012",
+    "IssueCredential": "didv:IssueCredential",
+    "LinkedDataSignature2015": "sec:LinkedDataSignature2015",
+    "LinkedDataSignature2016": "sec:LinkedDataSignature2016",
+    "RsaCryptographicKey": "sec:RsaCryptographicKey",
+    "RsaSignature2015": "sec:RsaSignature2015",
```
ottonomy: `RSASignature2017` is referenced in index.html section 6.5.1 and doesn't appear in this context.

kimdhamilton:
Ryan Grant: +1 to general direction. May have concerns on URNs; will follow up with details
@ChristopherA: +1 to direction

did-v1.jsonld
msporny: Yes, we're still finishing up implementations on that spec. We may want to back it out to RsaSignature2015, which does exist.

msporny: I just ended up adding RsaSignature2017 since we know we're implementing it.

PR #24, opened 2017-09-23
JSON object @context first line requirement

AxelNennker:
From the JSON spec "An object is an unordered set of name/value pairs". 
Only a serialization can have an ordering.

Could somebody please explain why "context" MUST be the first line?
If there is no real reason then I suggest to remove this requirement on DDOs.

cheers
Axel

### DIFF
--- a/did-v1.jsonld
+++ b/did-v1.jsonld
@@ -19,6 +19,7 @@
     "LinkedDataSignature2016": "sec:LinkedDataSignature2016",
     "RsaCryptographicKey": "sec:RsaCryptographicKey",
     "RsaSignature2015": "sec:RsaSignature2015",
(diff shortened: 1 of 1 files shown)


msporny:
You're right, Axel, @context being the first line isn't a requirement. We'll remove it.

AxelNennker:
The second commit removes the requirement that context must be the first line.

msporny:
LGTM, waiting on @talltree to approve the PR.

msporny:
Crap, sorry @AxelNennker I accidentally forced pushed something which deleted your changes when trying to pull your change set in... :(

I'll make the modifications and make you the committer in another change set...

AxelNennker:
Good. Thank you.

PR #31, opened 2017-10-28
Adapted method in the sovrin scheme to match the linked specification

retog:


msporny:
LGTM, this was an overzealous search/replace. Merging since it's purely editorial.

PR #34, opened 2017-11-30
Move standard key descriptions to separate specification.

msporny:
This PR splits the key descriptions out of the main DID spec.

### DIFF
--- a/ld-keys.html
+++ b/ld-keys.html
@@ -1,0 +1,199 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Linked Data Key Format Registry</title>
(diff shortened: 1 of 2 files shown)


ld-keys.html
```diff
@@ -0,0 +1,199 @@
...
+      <td>
+          Kim Duffy
+      </td>
+      <td>
+<pre>
+{
+  "id": "did:example:123456789abcdefghi/keys/2",
+  "type": ["CryptographicKey", "EdDsaSAKey"],
+  "owner": "did:example:123456789abcdefghi",
+  "curve": "ed25519",
```
dlongley: We should move "curve" into the key type.

msporny: Fixed.

PR #35, opened 2017-11-30
Fix JSON-LD context first line requirement.

msporny:
This is a long running PR that we need to pull in from @AxelNennker.


msporny:
Thanks for the PR @AxelNennker, sorry it took so long to pull it into the spec!

Issue #28, opened 2017-10-20
How are Service Endpoints defined?


kimdhamilton:
The current DID spec draft lists several examples of services, e.g. `socialWebInboxService` and `credentialRepositoryService`.

From what I can see:

1. In the DID LD-context, there are no terms related to services
2. The DID spec does not specify a data model for service endpoints, just examples like Extensible Data Interchange.

Is this correct? If so, then my current understanding of service endpoints is:

3. method specs define terms in their LD context
4. method specs define (or point to) the data model for service endpoints

Is this also correct?

msporny:
> In the DID LD-context, there are no terms related to services

Correct and we should decide if there are common services that do belong in there. For example, one could make an argument that credentialRepositoryService should go in there because a large subset of the population will have a Verifiable Claims Credential Repository.

> The DID spec does not specify a data model for service endpoints, just examples like Extensible Data Interchange.
> Is this correct? 

Correct, they're just nodes in the graph. The service endpoints themselves are probably specified in another specification (such as the Verifiable Claims spec).

> method specs define terms in their LD context

Service endpoints are most likely not method spec specific (some may be, but most shouldn't be).

> method specs define (or point to) the data model for service endpoints

Probably not, in general, you don't want to couple service endpoint descriptions with the DID or DID Method specs because that would put a constraint on innovation (requirement being that if you want your new service description in there, you MUST get it into the DID or DID method spec).

msporny:
I think we can at least say these things at present:

* Service Descriptions probably belong in their own specs, or sections of other specs (like Credentials Handler API spec).
* Terms for Service Descriptions may eventually find their way into the DID JSON-LD context if they are broadly supported, and until they do, they can be defined in a JSON-LD context that is not in the DID context nor a DID Method context.

We need more discussion in the group before we can make a decision one way or the other.

The other interesting data point here is that this may be a stronger argument for a "keys" and "services" container since we'd be able to keep the DID Context from growing uncontrollably based on the standardized services provided via DIDs.

msporny:
The spec now contains a section on service endpoints:

https://w3c-ccg.github.io/did-spec/#service-endpoints

Closing this issue.

PR #41, opened 2018-01-15
Update DID Spec to latest data model based on DID Hardening discussions

msporny:
This PR updates the spec to reflect the latest data model based on consensus agreement reached during the DID Hardening discussions.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/41.html" title="Last updated on Jan 25, 2018, 6:59 PM GMT (73b9a2d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/41/65f5638...73b9a2d.html" title="Last updated on Jan 25, 2018, 6:59 PM GMT (73b9a2d)">Diff</a>


index.html
```diff
@@ -2135,28 +2081,63 @@ <h1>Proposed DID Method Specifications</h1>
...
+    // this key can be used to authenticate as DID ...9938
+    "type": "PseudonymousBiometricBasedAuthentication2018",
+    "publicKey": "#bio-1"
+  }],
+
+  // EXPERIMENTAL: Proposal for encrypting messages for DID
+  "encryption": [{
+    // this key can be used to encrypt messages for the DID ...9938
+    "type": "RsaKeyBasedEncryption2018",
+    "publicKey": "#keys-2"
```
Fak3: #keys-2 is missing in the publicKey list

msporny: Thanks... fixing.

msporny: Fixed.

index.html
```diff
@@ -2040,92 +2041,37 @@ <h2>Alternate Serializations and Graph Models</h2>
...
 
     <tr>
       <td>
-          did:btcr:
-      </td>
-      <td>
-          Bitcoin
-      </td>
-      <td>
-          Christopher Allen
```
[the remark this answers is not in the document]
msporny: Updated in the DID Method Registry: https://github.com/w3c-ccg/did-method-registry/commit/cff6d66cccaa5cb16eb4a5a8c573602b74fd8e9b

peacekeeper:
This looks quite good to me, but some more work needs to be done on service endpoints:

1. In section 1.3.1. a "service" array is used, which I think was indeed the consensus during the DID Hardening/Harmonization process. But other sections such as 1.3.2. still use terms like "exampleService" or "credentialRepositoryService".

2. In section 4.4. and others we should decide what we say about what services. This is something the community will take (much?) more time to reach consensus on. We have people working on authentication services, credential repository services, messaging services, social web inbox services, Sovrin agents, DIF hubs, XDI, and others. We should decide if we either include ALL of them as examples, or only talk about services in very generic and neutral ways that don't favor any specific use or approach. See also https://github.com/w3c-ccg/did-spec/issues/28

I could volunteer to work on these two items?

msporny:
@peacekeeper wrote:
> I could volunteer to work on these two items?

Yes, please! :)

msporny:
Note that I intend to pull @peacekeeper's changes into this PR before merging: https://github.com/w3c-ccg/did-spec/pull/43

did-v1.jsonld
```diff
@@ -18,10 +18,13 @@
     "LinkedDataSignature2015": "sec:LinkedDataSignature2015",
     "LinkedDataSignature2016": "sec:LinkedDataSignature2016",
     "RsaCryptographicKey": "sec:RsaCryptographicKey",
+    "RsaKeyBasedAuthentication2018": "sec:RsaKeyBasedAuthentication",
```
dlongley: Seems like `KeyBased` is unnecessary -- just `RsaAuthentication2018` would do or `RsaSignatureAuthentication2018` would be more accurate?

did-v1.jsonld
```diff
@@ -18,10 +18,13 @@
...
+    "RsaSigningKey2018": "sec:RsaSigningKey",
     "RsaSignature2015": "sec:RsaSignature2015",
     "RsaSignature2017": "sec:RsaSignature2017",
     "UpdateDidDescription": "didv:UpdateDidDescription",
 
+    "authentication": "sec:authenticationSuite",
```
dlongley: I think this is "ok" ... I wonder if it's not actually the suite itself, but some "instantiation" of it + a particular key (just looking at the data). Maybe this is splitting hairs -- just expressing a little apprehension on this being slightly off.

msporny: Yes, you're correct. Will think about the proper term to use in the next set of edits.

index.html
```diff
@@ -584,7 +584,7 @@ <h2>Example Delegate-Managed DID Document</h2>
 </p>
 
 </section>
-</section>
+</section-->
```
peacekeeper: This comments out TWO `</section>` tags instead of one and messes up the numbering. Should be one line above.

index.html
```diff
@@ -513,26 +513,27 @@ <h2>Self-Managed DID Document</h2>
...
     "id": "did:example:123456789abcdefghi#keys-1",
-    "type": "RsaCryptographicKey",
+    "type": "RsaSigningKey2018",
     "owner": "did:example:123456789abcdefghi",
     "publicKeyPem": "-----BEGIN PUBLIC KEY...END PUBLIC KEY-----\r\n"
+  }],
+  "authentication": [{
+    // this key can be used to authenticate as DID ...9938
+    "type": "RsaSignatureAuthentication2018",
+    "publicKey": "#keys-1"
```
dlongley: We may want to support the full ID for the key here, not just the fragment identifier:

```js
"publicKey": "did:example:123456789abcdefghi#keys-1"
```

msporny:
Merging based on CCG consensus: https://w3c-ccg.github.io/meetings/2018-01-30/#resolution-1

mikelodder7:
In the Authentication section you have "authenticationCredential" but the example shows "authentication". I thought the consensus is to use "authentication" right?

mikelodder7:
In the services section you have in the example "service" when it should be "services" to be consistent with the notes above it.

msporny:
@mikelodder7 Correct, and fixed here: https://github.com/w3c-ccg/did-spec/commit/684e5f828353c65d875dd5fe60a17a286af7ad83#diff-eacf331f0ffc35d4b482f1d15a887d3bL563

msporny:
@mikelodder7 wrote:
> In the services section you have in the example "service" when it should be "services" to be consistent with the notes above it.

We should discuss on a call. When I switched to plural form, @dlongley noted how this backfired on the Web Payments WG, so we may just want to use singular form everywhere. In any case, we should get consensus on this in the group on singular vs. plural form.

mikelodder7:
I'm okay either way, but the we need the spec to be consistent, so as soon as we decide we should remember to update this to reflect the decision.

msporny:
@mikelodder7 wrote:
> as soon as we decide we should remember to update this to reflect the decision.

Raised an issue to remind us:

https://github.com/w3c-ccg/did-spec/issues/46

Thanks @mikelodder7 :)

PR #43, opened 2018-01-21
Update data model and examples for service endpoints.

peacekeeper:
***
<a href="https://pr-preview.s3.amazonaws.com/peacekeeper/did-spec/pull/43.html" title="Last updated on Jan 22, 2018, 9:02 AM GMT (698577a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/43/c6db2cb...peacekeeper:698577a.html" title="Last updated on Jan 22, 2018, 9:02 AM GMT (698577a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -525,7 +525,7 @@ when possible to express authorization to modify a DID Document.
     "publicKey": "#keys-1"
   }],
   "service": [{
-    "type": "example",
+    "type": "ExampleService",
     "endpoint": "https://example.com/endpoint/8377464"
   }]
 }
@@ -569,7 +569,7 @@ private keys.
     "biometricTemplateShard": "Mjk4MzQyO...5Mzg0MDI5Mwo="
   }],
   "service": [{
-    "type": "example",
+    "type": "ExampleService",
     "endpoint": "https://example.com/endpoint/8377464"
   }]
 }
@@ -984,7 +984,7 @@ Example:
     "biometricTemplateShard": "Mjk4MzQyO...5Mzg0MDI5Mwo="
   }],
   "service": [{
-    "type": "example",
+    "type": "ExampleService",
     "endpoint": "https://example.com/endpoint/8377464"
   }]
 }
@@ -1273,25 +1273,25 @@ JSON-LD object.
 <pre class="example nohighlight" title="Various service endpoints">
 {
   "service": [{
-    "type": "openid-connect-v3.1",
+    "type": "OpenIdConnectVersion3.1Service",
     "endpoint": "https://openid.example.com/"
   }, {
-    "type": "credential-repository",
+    "type": "CredentialRepositoryService",
     "endpoint": "https://repository.example.com/service/8377464"
   }, {
-    "type": "xdi",
+    "type": "XdiService",
     "endpoint": "https://xdi.example.com/8377464"
   }, {
-    "type": "agent",
+    "type": "AgentService",
     "endpoint": "https://agent.example.com/8377464"
   }, {
-    "type": "hub",
+    "type": "HubService",
     "endpoint": "https://hub.example.com/.identity/did:example:0123456789abcdef/"
   }, {
-    "type": "messaging",
+    "type": "MessagingService",
     "endpoint": "https://example.com/messages/8377464"
   }, {
-    "type": "social-web-inbox",
+    "type": "SocialWebInboxService",
     "endpoint": "https://social.example.com/83hfh37dj",
     "description": "My public social inbox",
     "spamCost": {
@@ -1299,7 +1299,7 @@ JSON-LD object.
       "currency": "USD"
     }
   }, {
-    "type": "did-auth-push-mode-v1",
+    "type": "DidAuthPushModeVersion1",
     "endpoint": "http://auth.example.com/did:example:123456789abcdefg"
   }]
 }
--- a/index.html
+++ b/index.html
@@ -565,7 +565,7 @@ private keys.
     "id": "did:example:123456789abcdefghi/biometric/1",
     "type": "PseudonymousBiometricTemplate2017",
     "owner": "did:example:123456789abcdefghi",
-    "biometricService": "https://example.com/authenticate"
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -525,8 +525,8 @@ <h2>Self-Managed DID Document</h2>
     "publicKey": "#keys-1"
   }],
   "service": [{
-    "type": "MessagingService",
-    "serviceEndpoint": "https://example.com/messages/8377464"
+    "type": "example",
```
msporny: A JSON-LD @type typically starts with an upper case letter. Suggest changing this to "ExampleService".

index.html
```diff
@@ -525,8 +525,8 @@ <h2>Self-Managed DID Document</h2>
...
+    "endpoint": "https://example.com/endpoint/8377464"
```
msporny: We need to be careful about simple names that might be stomped on when JSON-LD contexts are composed together. "serviceEndpoint" is more likely to survive naming conflicts than "endpoint". We may not be consistent in the rest of the spec w/ this point, in which case we should update other items that are likely to conflict.

index.html
```diff
@@ -560,14 +560,17 @@ <h2>Example Delegate-Managed DID Document</h2>
...
   "authenticationCredential": [{
     // this biometric can be used to authenticate as DID ...fghi
     "id": "did:example:123456789abcdefghi/biometric/1",
     "type": "PseudonymousBiometricTemplate2017",
     "owner": "did:example:123456789abcdefghi",
     "biometricService": "https://example.com/authenticate"
     "biometricTemplateShard": "Mjk4MzQyO...5Mzg0MDI5Mwo="
+  }],
+  "service": [{
+    "type": "example",
```
msporny: Again, "ExampleService", and "serviceEndpoint".

index.html
```diff
@@ -1256,17 +1270,38 @@ <h2>Service Endpoints</h2>
...
 </p>
 
-<pre class="example nohighlight" title="Service description for a Social Web inbox">
+<pre class="example nohighlight" title="Various service endpoints">
 {
-  "credentialRepositoryService": "https://repository.example.com/123",
-  "socialWebInboxService": {
-    "id": "https://social.example.com/83hfh37dj",
+  "service": [{
+    "type": "openid-connect-v3.1",
```
msporny: Should be "OpenIdConnectVersion3.1Service" or something similar, dashes are not typically used for property/type names.

index.html
```diff
@@ -1256,17 +1270,38 @@ <h2>Service Endpoints</h2>
...
+    "endpoint": "https://openid.example.com/"
+  }, {
+    "type": "credential-repository",
```
msporny: Should be "CredentialRepositoryService"

index.html
```diff
@@ -1256,17 +1270,38 @@ <h2>Service Endpoints</h2>
...
+    "endpoint": "https://repository.example.com/service/8377464"
+  }, {
+    "type": "xdi",
```
msporny: XdiService

index.html
```diff
@@ -1256,17 +1270,38 @@ <h2>Service Endpoints</h2>
...
+    "endpoint": "https://xdi.example.com/8377464"
+  }, {
+    "type": "agent",
+    "endpoint": "https://agent.example.com/8377464"
+  }, {
+    "type": "hub",
```
msporny: HubService (or something similar)... granted, "Hub" is probably too vague, may want to get more specific.

index.html
```diff
@@ -1256,17 +1270,38 @@ <h2>Service Endpoints</h2>
...
+    "type": "openid-connect-v3.1",
+    "endpoint": "https://openid.example.com/"
+  }, {
+    "type": "credential-repository",
+    "endpoint": "https://repository.example.com/service/8377464"
+  }, {
+    "type": "xdi",
+    "endpoint": "https://xdi.example.com/8377464"
+  }, {
+    "type": "agent",
```
msporny: "AgentService", but that's way too generic... should get far more specific.

index.html
```diff
@@ -1256,17 +1270,38 @@ <h2>Service Endpoints</h2>
...
+    "endpoint": "https://agent.example.com/8377464"
+  }, {
+    "type": "hub",
+    "endpoint": "https://hub.example.com/.identity/did:example:0123456789abcdef/"
+  }, {
+    "type": "messaging",
```
msporny: "MessagingService"... again, should probably be something like "LinkedDataNotificationsService", pointing to a specific spec name.

index.html
```diff
@@ -2110,9 +2110,9 @@ <h1>Real World Example</h1>
     "id": "did:example:123456789abcdefghi#bio-1",
     "type": "PseudonymousBiometricTemplate2017",
     "owner": "did:example:123456789abcdefghi",
-    "biometricService": "https://example.com/authenticate"
+    "biometricService": "https://example.com/authenticate",
```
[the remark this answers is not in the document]
peacekeeper: Not suggesting to remove the line, just adding a missing `,` comma at the end. Actually now that I see it, I need to fix this in other places too.

index.html
peacekeeper: Okay I'll change this everywhere, but lower case is used in both the Hardening and Harmonization docs I think? This may be again a question of "RDF worldview" vs "JSON worldview"...

index.html
peacekeeper: I agree, the term "agent" seems to be sometimes in a broad sense and sometimes for a specific protocol.

index.html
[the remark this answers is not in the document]
msporny: Ah, sorry, missed that. Wondering if we should change "Service" to something else so people don't get confused about this being a service... it was really meant to be "here's the server that can tell you if an image matches the template". Fine keeping it, or you can suggest something else.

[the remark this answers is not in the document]
peacekeeper: I think you understand DIDs+biometrics better than most, so whatever makes sense to you should be fine..

PR #48, opened 2018-02-06
fix references and examples

zoellner:
a couple of references were outdated due to the removal of sections. some examples were outdated and didn't match the specification.

***
<a href="https://pr-preview.s3.amazonaws.com/zoellner/did-spec/pull/48.html" title="Last updated on Feb 6, 2018, 6:44 PM GMT (eda9a2d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/48/4e6c77c...zoellner:eda9a2d.html" title="Last updated on Feb 6, 2018, 6:44 PM GMT (eda9a2d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1413,11 +1413,11 @@ integrity of the DID Document according to either:
 </p>
 <ol start="1">
   <li>
-The entity as defined in section 6.4, or if not present:
+The entity as defined in section 4.4, or if not present:
   </li>
 
   <li>
-The delegate as defined in section 6.3.
+The delegate as defined in section 4.3.
   </li>
 </ol>
 
--- a/index.html
+++ b/index.html
@@ -2151,16 +2151,20 @@ A future-facing real-world context is provided below:
   }],
 
   "authentication": [{
-    // this key can be used to authenticate as DID ...9938
+    // this key can be used to authenticate as DID ...fghi
+    "id": "did:example:123456789abcdefghi#keys-1"
     "type": "RsaSignatureAuthentication2018",
+    "owner": "did:example:123456789abcdefghi"
     "publicKey": "did:example:123456789abcdefghi#keys-1"
   }, {
-    // this key can be used to authenticate as DID ...9938
+    // this key can be used to authenticate as DID ...fghi
+    "id": "did:example:123456789abcdefghi#bio-1"
     "type": "PseudonymousBiometricAuthentication2018",
+    "owner": "did:example:123456789abcdefghi"
     "biometricTemplate": "did:example:123456789abcdefghi#bio-1"
   }],
 
-  "service": [{
+  "services": [{
     "type": "OpenIdConnectVersion3.1Service",
     "serviceEndpoint": "https://openid.example.com/"
   }, {
--- a/index.html
+++ b/index.html
@@ -1384,7 +1384,7 @@ The key for this property MUST be updated.
 
   <li>
 The value of this key MUST follow the formatting rules (3, 4, 5)
-from section 6.7.
+from section 4.5.
   </li>
 </ol>
 
--- a/index.html
+++ b/index.html
@@ -1268,7 +1268,7 @@ JSON-LD object.
 
 <pre class="example nohighlight" title="Various service endpoints">
 {
-  "service": [{
+  "services": [{
     "type": "OpenIdConnectVersion3.1Service",
     "serviceEndpoint": "https://openid.example.com/"
   }, {
--- a/index.html
+++ b/index.html
@@ -562,9 +562,7 @@ private keys.
   }],
   "authentication": [{
     // this biometric can be used to authenticate as DID ...fghi
-    "id": "did:example:123456789abcdefghi/biometric/1",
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1046,12 +1046,16 @@ <h2>Authentication</h2>
   "id": "did:example:123456789abcdefghi",
   ...
   "authentication": [{
-    // this key can be used to authenticate as DID ...9938
+    // this key can be used to authenticate as DID ...fghi
+    "id": "did:example:123456789abcdefghi#keys-1"
```
msporny: This is a link to the key, which is wrong. The object being edited is an authentication suite, not a key object. The authentication suite doesn't have an "id".

index.html
```diff
@@ -1046,12 +1046,16 @@ <h2>Authentication</h2>
...
     "type": "RsaSignatureAuthentication2018",
+    "owner": "did:example:123456789abcdefghi"
```
msporny: authentication suites do not have owners.

index.html
```diff
@@ -1046,12 +1046,16 @@ <h2>Authentication</h2>
...
     "publicKey": "did:example:123456789abcdefghi#keys-1"
   }, {
-    // this key can be used to authenticate as DID ...9938
+    // this key can be used to authenticate as DID ...fghi
+    "id": "did:example:123456789abcdefghi#bio-1"
```
msporny: authentication suites do not have ids

index.html
```diff
@@ -1046,12 +1046,16 @@ <h2>Authentication</h2>
...
     "type": "PseudonymousBiometricAuthentication2018",
+    "owner": "did:example:123456789abcdefghi"
```
msporny: authentication suites do not have owners.

index.html
```diff
@@ -1384,7 +1388,7 @@ <h2>Updated (Optional)</h2>
 
   <li>
 The value of this key MUST follow the formatting rules (3, 4, 5)
-from section 6.7.
+from section 4.5.
```
msporny: While this is accurate, it's just going to get messed up again unless we use ReSpec's <a href="#frag-id"></a> format. We should update to use that pattern everywhere.

index.html
```diff
@@ -1592,7 +1596,7 @@ <h1>Security Considerations</h1>
       As the root identifier records for decentralized identifiers, DIDs and DID Documents are a vital component of decentralized identity management. They are also the foundational building blocks of DPKI (decentralized public key infrastructure) as an augmentation to conventional X.509 certificates.
 
 
-      As such, DIDs are designed to operate under the general Internet threat model used by many IETF standards. We assume uncompromised endpoints, but allow messages to be read or corrupted on the network. Protecting against an attack when a system is compromised requires external key-signing hardware. See also section 6.5 regarding key revocation and recovery.
+      As such, DIDs are designed to operate under the general Internet threat model used by many IETF standards. We assume uncompromised endpoints, but allow messages to be read or corrupted on the network. Protecting against an attack when a system is compromised requires external key-signing hardware. See also section <a href="#key-revocation-and-recover"></a> regarding key revocation and recovery.
```
msporny: Yes, this form is correct... but the frag id is wrong... it should be: "#key-revocation-and-recovery"

index.html
```diff
@@ -2147,16 +2151,20 @@ <h1>Real World Example</h1>
   }],
 
   "authentication": [{
-    // this key can be used to authenticate as DID ...9938
+    // this key can be used to authenticate as DID ...fghi
+    "id": "did:example:123456789abcdefghi#keys-1"
```
msporny: Same issue w/ the fixes here... the committer assumed these were key objects, which they are not. They are authentication suite objects. This is probably also a lesson for the CG, that developers will just assume that key information goes in here... I expect many libraries will throw an error, but we'll probably have to include spec text that ensures that they throw on stuff that isn't properly formed.

index.html
zoellner: updated to use correct frag id. updated existing links to wrong frag id as well

index.html
zoellner: updated to use ReSpec

index.html
zoellner: There must be a misunderstanding on my side then about rules 2 and 3 under 4.3 Authentication which states: 
> 2. The value of the authentication property should be an array of proof mechanisms.
> 3. Each proof mechanism must include id, type, and owner properties.

The example in the currently commented out section "Example Delegate-Managed DID Document" has this format as well.

Can you clarify?

msporny: >  id, type, and owner properties

Ah, you have spotted two bugs in the spec. Thank you! Neither id or owner is required. This is old spec text from when we used to put key information in "authentication".

zoellner: I've removed the id and owner properties. I leave it up to you to fix the spec since I want to keep this PR as editorial.

zoellner:
IPR release is signed and all requested changes are addressed
Thank you for the help @msporny

msporny:
LGTM, editorial, squashed and merged to keep history cleaner. Thank you @zoellner!

PR #50, opened 2018-02-15
 Add 'Public Keys' section and update 'Authentication' section.

peacekeeper:
***
<a href="https://pr-preview.s3.amazonaws.com/peacekeeper/did-spec/pull/50.html" title="Last updated on Feb 25, 2018, 8:33 AM GMT (c5f1528)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/50/1e0ecb5...peacekeeper:c5f1528.html" title="Last updated on Feb 25, 2018, 8:33 AM GMT (c5f1528)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -988,6 +988,100 @@ Example:
 </section -->
 
 <section>
+<h2>Public Keys</h2>
+
+<p>
+Public keys are used for digital signatures, encryption and other cryptographic operations, 
+which in turn are the basis for purposes such as authentication
+(see Section <a href="#authentication"></a>)
+or establishing secure communication with <a>service endpoints</a>
+(see Section <a href="#service-endpoints"></a>).
+In addition, public keys may play a role in authorization mechanisms of DID CRUD operations 
+(see Section <a href="#did-operations"></a>); This
+may be defined by DID Method specifications.
+</p>
+
+<p>
+The primary intention is that a DID Document lists public keys whose corresponding
+private keys are controlled by the entity identified by the DID ("owned" public keys).
+However, a DID Document MAY also list "non-owned" public keys.
+</p>
+
+<p>
+The rules for public keys are:
+</p>
+
+<ol start="1">
+  <li>
+A DID Document MAY include a <code>publicKey</code>
+property.
+  </li>
+
+  <li>
+The value of the <code>publicKey</code> property should be an
+array of public keys.
+  </li>
+
+  <li>
+Each public key must include <code>id</code> and
+<code>type</code> properties, and exactly one value property.
+  </li>
+
+  <li>
+Each public key may include an <code>owner</code> property, which identifies the entity that controls
+the corresponding private key. If this property is missing, it is assumed to be the DID subject.
+  </li>
+
+  <li>
+The value property of a public key may be
+<code>publicKeyPem</code>, <code>publicKeyJwk</code>, <code>publicKeyHex</code>, <code>publicKeyBase64</code> or similar,
+depending on the format and encoding of the public key.
+  </li>
+
+  <li>
+A registry of key types and formats is available at [[LINKED-DATA-KEY-FORMAT-REGISTRY]].
+  </li>
+</ol>
+
+<p>
+Example:
+</p>
+
+<pre class="example nohighlight"
+  title="Various public keys">
+{
+  "@context": ["https://w3id.org/did/v1", "https://w3id.org/security/v1"],
+  "id": "did:example:123456789abcdefghi",
+  ...
+  "publicKey": [{
+    "id": "did:example:123456789abcdefghi#keys-1",
+    "type": "RsaSigningKey",
+    "owner": "did:example:123456789abcdefghi",
+    "publicKeyPem": "-----BEGIN PUBLIC KEY...END PUBLIC KEY-----\r\n"
+  }, {
+    "id": "did:example:123456789abcdefghi#keys-2",
+    "type": "Ed25519VerificationKey",
+    "owner": "did:example:123456789abcdefghi",
+    "publicKeyBase58": "H3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV"
+  }, {
+    "id": "did:example:123456789abcdefghi#keys-3",
+    "type": "EdDsaSAPublicKeySecp256k1",
+    "owner": "did:example:123456789abcdefghi",
+    "publicKeyHex": "02b97c30de767f084ce3080168ee293053ba33b235d7116a3263d29f1450936b71"
+  }],
+  ...
+}
+</pre>
+
+<p class="note">
+Caching and expiration of the keys in a DID Document is entirely the
+responsibility of DID resolvers and other clients. See Section
+<a href="#did-resolvers"></a>.
+</p>
+
+</section>
+
+<section>
 <h2>Authentication</h2>
 
 <p>
@@ -1053,12 +1147,6 @@ Example:
   ...
 }
 </pre>
-
-<p class="note">
-Caching and expiration of the keys in a DID Document is entirely the
-responsibility of DID resolvers and other clients. See Section
-<a href="#did-resolvers"></a>.
-</p>
 
 </section>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -987,6 +987,100 @@ <h2>Delegates</h2>
...
+
+<pre class="example nohighlight"
+  title="Various public keys">
+{
+  "@context": ["https://w3id.org/did/v1", "https://w3id.org/security/v1"],
+  "id": "did:example:123456789abcdefghi",
+  ...
+  "publicKey": [{
+    "id": "did:example:123456789abcdefghi#keys-1",
+    "type": "RsaSigningKey",
```
msporny: Should be "RsaSigningKey2018"

index.html
```diff
@@ -987,6 +987,100 @@ <h2>Delegates</h2>
...
+    "owner": "did:example:123456789abcdefghi",
+    "publicKeyPem": "-----BEGIN PUBLIC KEY...END PUBLIC KEY-----\r\n"
+  }, {
+    "id": "did:example:123456789abcdefghi#keys-2",
+    "type": "Ed25519VerificationKey",
```
msporny: Should be "EdwardsVerificationKey2018" -- we need to figure out if we're going to use "Signing" or "Verification". I'm fine committing something that's schizophrenic for now for force the discussion later on.

index.html
```diff
@@ -987,6 +987,100 @@ <h2>Delegates</h2>
...
+    "owner": "did:example:123456789abcdefghi",
+    "publicKeyBase58": "H3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV"
+  }, {
+    "id": "did:example:123456789abcdefghi#keys-3",
+    "type": "EdDsaSAPublicKeySecp256k1",
```
msporny: Crypto suites should have dates on them as it's easier for developers to figure out if they're using dated software/signatures/etc. or not.

index.html
```diff
@@ -987,6 +987,100 @@ <h2>Delegates</h2>
...
+    "owner": "did:example:123456789abcdefghi",
+    "publicKeyHex": "02b97c30de767f084ce3080168ee293053ba33b235d7116a3263d29f1450936b71"
```
msporny: We need to make sure this is the right thing to do. Feels bloated and unwieldy. If these are raw bytes, we need to understand why we're not using a better encoding scheme (base64, base58, etc.).

index.html
```diff
@@ -1019,15 +1113,13 @@ <h2>Authentication</h2>
...
+Each proof mechanism must include the
+<code>type</code> property.
   </li>
 
   <li>
-Each proof mechanism MAY be a key description of a valid public key or
-verification key. A list of standard key descriptions is included in
-Appendix <a href="#standard-key-descriptions"></a>.
-A new key description MAY also be defined by a DID method specification.
+Each proof mechanism MAY reference a public key or verification
```
msporny: This is not what the proposal was. Each proof mechanism MAY embed the public key or MAY a public key.

We should also decide what we're doing wrt. language around "public key" "signing key", "verification key", we should settle on something consistent.

index.html
dlongley: I think it should use `Ed25519` rather than `Edwards` as a prefix because that's what developers will be looking for (IMO). But I do agree that we need to figure out if we're using "Verification" or "Signing" or "SignatureVerification" or something. My problem with "Verification" is that it makes no statement about "what" is being verified. Getting the term "signature" in there is helpful, but perhaps briefly exploring what other types of keys there might be would also make it more obvious what to choose here. I think the most prominent example is Signing vs. Encryption. If those are the only possibilities, then perhaps "xVerificationKey" is ok -- because it would be "xEncryptionKey" for a public key used to encrypt.

mikelodder7: Elliptic Curves are not used for encryption/decryption. Just key exchange and signing.
I agree it should be Ed25519. That is how it is most commonly known. I have not met anyone that refers to it as Edwards. Twisted Edwards are a class of elliptic curves. If we did that you would have to start calling other curves by their full class names like Weierstrass or Montgomery or Jacobian. Lets stick to the actual curve names.

SignatureVerification while that label is clear feels long. In cryptography, "Verification" only refers to verifying a signature with Ed25519. I don't have a strong opinion here so I'm okay with SignatureVerification.

index.html
mikelodder7: EdDsaSAPublicKeySecp256k1 is incorrect. EdDSA is only possible with Twisted Edward curves like Ed25519.
ECDSA is for linear curves like Montgomery and Weierstrass curves. If the curve is Secp256k1 then the name should be 

Secp256k1SignatureVerification or Secp256k1ECDSA. I prefer the first due to the discussion above. SignatureVerification describes what the key is used for and ECDSA is the algorithm to perform the signature verification.

index.html
mikelodder7: Can we also be clear that signing keys should never be public?
Lets not use "signing key" anywhere as it could encourage bad behavior.
The only keys that DID's should be including are public keys that encrypt (as with RSA), verify signatures, or key exchange (without forward secrecy).
That being said, the only keys I would recommend in DIDs would be SignatureVerification keys. I can be convinced otherwise with a good use case. If these are not crypto keys then we should designate them so with labels like "Crypto" or "Cryptographic" so other keys can use terms like "Biometric".

index.html
msporny: +1 to Secp256k1SignatureVerification

index.html
msporny: > Elliptic Curves are not used for encryption/decryption. 

Pushing back a bit on this. You're technically correct, common usage is symmetric keys are used w/ Elliptic Curve public keys to do encryption/decryption... but we are defining a public key here, right? So, given something like ECIES we could say this is "a public key that is used for ECIES-style encryption, where the symmetric key is just a part of the protocol". Wouldn't that mean that we may have "Ed25519EncryptingKey"... which means "if you want to send something to me in encrypted form, use this key".

index.html
msporny: > Can we also be clear that signing keys should never be public?

+1, we should add some spec text in the DID spec for that. Although, I'd hope that developers writing applications that deal with keys would know that.

> Lets not use "signing key" anywhere as it could encourage bad behavior.

Ok, so "SignatureVerificationKey"? or just "VerificationKey"? I think I prefer something like: "RsaVerificationKey2018", "Ed25519VerificationKey2018", "Secp256k1VerificationKey2018"... thoughts?

> The only keys that DID's should be including are public keys that encrypt (as with RSA), verify signatures, or key exchange (without forward secrecy).

+1, we should make this clear in the spec.

> That being said, the only keys I would recommend in DIDs would be SignatureVerification keys.

My gut says that's probably true, and we'd say something like "BiometricTemplateId" for public pseudonymous biometric template IDs (clearly, it's a bad idea to put any raw or encrypted biometric data on a ledger). I'm hesitant to say MUST in the spec... we should warn folks in the spec about placing sensitive crypto material in a ledger, even in encrypted form.

index.html
mikelodder7: Yes you could use it that way but its misleading because what algorithm is used for symmetric encryption? Is it AES (hopefully, but not always the case where encrypting an integer less than 128 bits long can bloat a database column and Blowfish is used)? Are you just XOR'ing shared material (Bad idea)? ECIES is not clear enough by itself. Unfortunately I don't have a good recommendation for this right now but we should have a separate discussion on the naming convention.

index.html
mikelodder7: --  Although, I'd hope that developers writing applications that deal with keys would know that.

I hope so too. But it couldn't hurt to direct developers.

peacekeeper:
Changed key terms now to `RsaSignatureVerificationKey2018`, `Ed25519SignatureVerificationKey2018`, `Secp256k1SignatureVerificationKey2018`. Looks a bit long to me but clear and accurate. Everyone okay with that?

msporny:
> Looks a bit long to me but clear and accurate. 

If a developer is using JSON-LD, they can compact to shorter names of their choosing. JSON-LD developers w/ standard tooling can compact to something like "Secp256k1Svk2018" if they want... or "BitcoinKey2018"... or whatever. 

Granted, that doesn't help JSON developers, who will just have to live with the names we pick.

peacekeeper:
Fixes #52

msporny:
LGTM. We may want to reduce to "RsaVerificationKey2018" per @mikelodder7's comments. If there is nothing that VerificationKey can be confused with, then we should shorten it.

peacekeeper:
Changed key terms now to shorter `RsaVerificationKey2018`, `Ed25519VerificationKey2018`, `Secp256k1VerificationKey2018`. Any more comments or should we merge?

msporny:
LGTM, merging.

PR #61, opened 2018-03-09
Clarify PublicKey Section

AxelNennker:
Still needs work e.g. the registry is incomplete. I picked one format as MTI (discuss)

***
<a href="https://pr-preview.s3.amazonaws.com/AxelNennker/did-spec/pull/61.html" title="Last updated on Apr 3, 2018, 7:57 AM GMT (b9eb63c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/61/8ea8c86...AxelNennker:b9eb63c.html" title="Last updated on Apr 3, 2018, 7:57 AM GMT (b9eb63c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1034,7 +1034,7 @@ the corresponding private key. If this property is missing, it is assumed to be 
 
   <li>
 The value property of a public key MUST be exactly one of
-<code>publicKeyPem</code>, <code>publicKeyJwk</code>, <code>publicKeyHex</code>, <code>publicKeyBase64</code>,
+<code>publicKeyPem</code>, <code>publicKeyJwk</code>, <code>publicKeyHex</code>, <code>publicKeyBase64</code>, <code>publicKeyBase58</code>
 depending on the format and encoding of the public key. A conforming implementation of this specification MUST support <code>publicKeyPem</code>.
   </li>
 
--- a/index.html
+++ b/index.html
@@ -1017,24 +1017,25 @@ property.
   </li>
 
   <li>
-The value of the <code>publicKey</code> property should be an
-array of public keys.
+The value of the <code>publicKey</code> property SHOULD be an
+array of public keys. If the value is not an array then it MUST be exactly one public key. 
   </li>
 
   <li>
-Each public key must include <code>id</code> and
-<code>type</code> properties, and exactly one value property.
+Each public key MUST include <code>id</code> and
+<code>type</code> properties, and exactly one value property. The array of public keys SHOULD not contain dublicate entries
+    with the same id and different value properties with different formats.
   </li>
 
   <li>
-Each public key may include an <code>owner</code> property, which identifies the entity that controls
+Each public key MAY include an <code>owner</code> property, which identifies the entity that controls
 the corresponding private key. If this property is missing, it is assumed to be the DID subject.
   </li>
 
   <li>
-The value property of a public key may be
-<code>publicKeyPem</code>, <code>publicKeyJwk</code>, <code>publicKeyHex</code>, <code>publicKeyBase64</code> or similar,
-depending on the format and encoding of the public key.
+The value property of a public key MUST be exactly one of
+<code>publicKeyPem</code>, <code>publicKeyJwk</code>, <code>publicKeyHex</code>, <code>publicKeyBase64</code>,
+depending on the format and encoding of the public key. A conforming implementation of this specification MUST support <code>publicKeyPem</code>.
   </li>
 
   <li>
--- a/index.html
+++ b/index.html
@@ -1023,7 +1023,7 @@ array of public keys. If the value is not an array then it MUST be exactly one p
 
   <li>
 Each public key MUST include <code>id</code> and
-<code>type</code> properties, and exactly one value property. The array of public keys SHOULD not contain dublicate entries
+<code>type</code> properties, and exactly one value property. The array of public keys SHOULD not contain duplicate entries
     with the same id and different value properties with different formats.
   </li>
 
--- a/index.html
+++ b/index.html
@@ -1035,7 +1035,7 @@ the corresponding private key. If this property is missing, it is assumed to be 
   <li>
 The value property of a public key MUST be exactly one of
 <code>publicKeyPem</code>, <code>publicKeyJwk</code>, <code>publicKeyHex</code>, <code>publicKeyBase64</code>, <code>publicKeyBase58</code>
-depending on the format and encoding of the public key. A conforming implementation of this specification MUST support <code>publicKeyPem</code>.
+depending on the format and encoding of the public key. 
   </li>
 
   <li>
--- a/index.html
+++ b/index.html
@@ -1017,8 +1017,8 @@ property.
   </li>
 
   <li>
-The value of the <code>publicKey</code> property SHOULD be an
-array of public keys. If the value is not an array then it MUST be exactly one public key. 
+The value of the <code>publicKey</code> property MUST be an
+array of public keys. 
   </li>
 
   <li>


index.html
```diff
@@ -1017,24 +1017,25 @@ <h2>Public Keys</h2>
...
-array of public keys.
+The value of the <code>publicKey</code> property SHOULD be an
+array of public keys. If the value is not an array then it MUST be exactly one public key. 
   </li>
 
   <li>
-Each public key must include <code>id</code> and
-<code>type</code> properties, and exactly one value property.
+Each public key MUST include <code>id</code> and
+<code>type</code> properties, and exactly one value property. The array of public keys SHOULD not contain dublicate entries
```
peacekeeper: Typo: dublicate -> duplicate

AxelNennker:
@peacekeeper thanks

index.html
```diff
@@ -1017,24 +1017,25 @@ <h2>Public Keys</h2>
...
 the corresponding private key. If this property is missing, it is assumed to be the DID subject.
   </li>
 
   <li>
-The value property of a public key may be
-<code>publicKeyPem</code>, <code>publicKeyJwk</code>, <code>publicKeyHex</code>, <code>publicKeyBase64</code> or similar,
-depending on the format and encoding of the public key.
+The value property of a public key MUST be exactly one of
+<code>publicKeyPem</code>, <code>publicKeyJwk</code>, <code>publicKeyHex</code>, <code>publicKeyBase64</code>,
+depending on the format and encoding of the public key. A conforming implementation of this specification MUST support <code>publicKeyPem</code>.
```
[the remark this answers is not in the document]
davidlehn: Also why is publicKeyPem support a "MUST"?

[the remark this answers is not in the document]
AxelNennker: I picked one and that was publicKeyPem . I have no preference here. I all current implementations support all listed then I would leave the MTI part out. If there is a preference in the dev community or reasons to prefer one then demanding one to be mandatory helps interoperability. I hoped to start a discussion here. No offense.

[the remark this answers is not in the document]
AxelNennker: Regarding completeness of the list... The spec should not be open ended but suggest something common and reasonable, I think. 
I suggest to add publicKeyBase58 to the list and then run the new list/wording past the group.

index.html
```diff
@@ -1017,24 +1017,25 @@ <h2>Public Keys</h2>
...
 the corresponding private key. If this property is missing, it is assumed to be the DID subject.
   </li>
 
   <li>
-The value property of a public key may be
-<code>publicKeyPem</code>, <code>publicKeyJwk</code>, <code>publicKeyHex</code>, <code>publicKeyBase64</code> or similar,
-depending on the format and encoding of the public key.
+The value property of a public key MUST be exactly one of
+<code>publicKeyPem</code>, <code>publicKeyJwk</code>, <code>publicKeyHex</code>, <code>publicKeyBase64</code>, <code>publicKeyBase58</code>
+depending on the format and encoding of the public key. A conforming implementation of this specification MUST support <code>publicKeyPem</code>.
```
pelle: @AxelNennker why this requirement for supporting a particular implementation. I am only aware of 1 did method using publicKeyPem and most using hex or base64, so it doesn't seem to make sense

index.html
```diff
@@ -1017,24 +1017,25 @@ <h2>Public Keys</h2>
   </li>
 
   <li>
-The value of the <code>publicKey</code> property should be an
-array of public keys.
+The value of the <code>publicKey</code> property SHOULD be an
+array of public keys. If the value is not an array then it MUST be exactly one public key.
```
pelle: Why not just require it to be an array, this simplifies implementation and makes it less error prone

index.html
dlongley: I think `v1` and `sov` are using `publicKeyBase58`. Not sure about `btcr`.

AxelNennker: I just picked one and I don't care much which one is MTI. I think it is good for interoperability when a spec mandates at least one methods to be implemented - otherwise the intersection of implemented methods might be empty. Requiring all currently known methods might be a too high burden for implementations to be spec compliant. Let the community decide which is the one or just use publicKeyBase58

dlongley: @AxelNennker,

I understand where you're coming from and that you "just picked one". But I'm just not sure that the interoperability amongst DID methods occurs at the level of public key encoding.

I'm trying to envision the test that would be written to confirm interop here -- and it would perhaps involve a signature verification application that can verify two different signatures. One signature would be verified using a key from DID Document D1 on method M1's ledger and another by using a key from DID Document D2 on method M2's ledger. It wouldn't matter which encoding was used for the keys, but rather that the verifier could find the keys the same way in either DID Document from either method.

Having only one type of key encoding would certainly simplify the verifier, but it seems, at least to me, that the level of interop we're looking for here is more in how the appropriate key is found (and that it's found the same way). I don't see a reason to impose a restriction that a certain encoding must be provided (i.e. "just pick one" like you've suggested here) -- at least not at the DID spec level. That seems to be an important requirement, perhaps, at the signature suite level.

index.html
AxelNennker: Allowing the same public key to be specified in more than one format might lead to attacks where the key is encoded in one format but the format specified is different. So usually a protocol is harder to attack when there is not much to negotiate and everything is agreed upon off-channel and a-priory.
So restricting formats up front is a security feature and a interoperability feature.
I want this spec to move forward and usually it is good in standardization to have ONE WAY to do something and not two or more. Moving the important bits into other specs (signature suites) makes this specification harder to read and to implement. I prefer one document that a developer can read from the beginning to the end and implement it without having to read several other documents - sometime this can't be avoided when other well-established standards are referenced.

msporny:
I'm trying to figure out how to pull this PR in, but it contains changes that I would expect would create objections (like requiring that publicKeyPem or base58 key encoding formats are ALWAYS supported). I guess I'll start a review and mark the items that are controversial but I'm afraid that after I'm done, there won't be much left. 

Just trying to move this PR along.

index.html
msporny: Hmm, I wasn't reading your change in that way @AxelNennker. While I agree with what you're saying, your spec text doesn't have anything to do w/ using only one key format.

I'm fine w/ @pelle's restriction as it does help people that want to just process this information as JSON data (although I expect eventually those implementations will have to deal w/ developers mis-implementing and not putting items in an array). Yes, you can make JSON Schema enforce, but that assumes everyone will enforce that rule when in reality, some developers do sloppy implementations that then create backwards compatibility issues.

index.html
```diff
@@ -1017,24 +1017,25 @@ <h2>Public Keys</h2>
...
-array of public keys.
+The value of the <code>publicKey</code> property SHOULD be an
+array of public keys. If the value is not an array then it MUST be exactly one public key. 
   </li>
 
   <li>
-Each public key must include <code>id</code> and
-<code>type</code> properties, and exactly one value property.
+Each public key MUST include <code>id</code> and
+<code>type</code> properties, and exactly one value property. The array of public keys SHOULD not contain duplicate entries
```
msporny: What does "exactly one value property" mean? This text is vague.

index.html
```diff
@@ -1017,24 +1017,25 @@ <h2>Public Keys</h2>
...
+    with the same id and different value properties with different formats.
   </li>
 
   <li>
-Each public key may include an <code>owner</code> property, which identifies the entity that controls
+Each public key MAY include an <code>owner</code> property, which identifies the entity that controls
```
msporny: This is now a MUST after conversations during the telecons and RWoT. We have all implementers on board.

index.html
msporny: Yeah, I feel like it'll be nearly impossible to get the BTCR, uPort, and Sovrin implementers to support PEM-based encoding... or rather, any one blockchain implementation to support the other blockchain keys as THE one way to implement. This is not where we're looking for interoperability.

index.html
AxelNennker: You noticed that this PR did not change the text you call 'vague', right? I suggest to merge this and then improve the vague text.

rhiaro:
Closing, as approved changes from this PR are now pulled into https://github.com/w3c-ccg/did-spec/pull/162

PR #66, opened 2018-03-12
Resolvers MUST NOT return properties if signature validation fails

AxelNennker:
Resolvers MUST NOT return DID Document properties if signature validation fails

***
<a href="https://pr-preview.s3.amazonaws.com/AxelNennker/did-spec/pull/66.html" title="Last updated on Apr 3, 2018, 7:43 AM GMT (550ee69)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/66/b0f2168...AxelNennker:550ee69.html" title="Last updated on Apr 3, 2018, 7:43 AM GMT (550ee69)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1657,7 +1657,8 @@ is signed.
   </li>
 
   <li>
-MAY offer the service of returning requested properties of the DID Document.
+MAY offer the service of returning requested properties of the DID Document. If this service is offered and the DID Document is
+signed then property values MUST only be returned if the signature is valid - otherwise an error MUST be returned.
   </li>
 </ol>
 
--- a/index.html
+++ b/index.html
@@ -1660,6 +1660,9 @@ is signed.
 MAY offer the service of returning requested properties of the DID Document. If this service is offered and the DID Document is
 signed then property values MUST only be returned if the signature is valid - otherwise an error MUST be returned.
   </li>
+  <li>
+If this service is offered and the DID Document can not be verified per the rules in the corresponding DID Method Specification,
+then the Resolver MUST return an error.
 </ol>
 
 </section>
@@ -1682,6 +1685,9 @@ signed then property values MUST only be returned if the signature is valid - ot
       For their part, the DLTs hosting DIDs and DID Documents have special security properties for preventing active attacks. Their design uses public/private key cryptography to allow operation on passively monitored networks without risking compromise of private keys. This is what makes DID architecture and decentralized identity possible.
 
 
+      Validating the signature of the DID document and returning properties of the DID must ensure that the signature validating
+      algorithm and the service returning properties return exactly those properties that are covered by the valid signature.
+      Separating signature validation and property extraction repeatedly lead to security vulnerabilities in the past.
 </p>
 
 <section>


index.html
```diff
@@ -1657,7 +1657,8 @@ <h1>DID Resolvers</h1>
   </li>
 
   <li>
-MAY offer the service of returning requested properties of the DID Document.
+MAY offer the service of returning requested properties of the DID Document. If this service is offered and the DID Document is
```
msporny: I suggest we change this language to "If this service is offered and the entire DID Document can not be verified per the rules in the corresponding DID Method Specification, then the Resolver MUST return an error."

msporny: ... which raises another problem w/ the spec. We probably need a separate spec for resolvers. We should take all these resolver requirements out of the DID Spec and put them in a Resolver specification.

dlongley: Slight tweak to manu's suggestion, I think we can just say "DID Document" not "entire DID Document".

I also don't know if we need a separate spec for resolvers -- perhaps a separate section would suffice. It just depends on how much content there is, IMO.

msporny:
This PR is stuck at present, waiting for @AxelNennker to make changes or suggest alternatives.

AxelNennker:
Hi @msporny 
was surprised that this PR was stuck. I added your proposal to this PR and a security note regarding signature validation.

index.html
TomCJones: i am a bit lost on the purpose of this spec. Is the normative description of the resolve in scope or should that be in a separate document. If it is normative to the resolve, then that should be clearly stated somewhere.

peacekeeper:
This should be addressed by the DID Resolution spec: https://github.com/w3c-ccg/did-resolution/

mwherman2000:
> Resolvers MUST NOT return DID Document properties if signature validation fails
> 
> [Preview](https://pr-preview.s3.amazonaws.com/AxelNennker/did-spec/pull/66.html) | [Diff](https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/66/b0f2168...AxelNennker:550ee69.html)

Based on Example 4 in https://w3c-ccg.github.io/did-spec/#did-subject, is there a strict requirement for a DID Document to be signed to be considered a valid DID Document?

```
EXAMPLE 4
{
  "id": "did:example:21tDAKCERh95uGgKbJNHYp"
}
```

peacekeeper:
@mwherman2000 There is no requirement for a DID Document to be signed, and it is very important to understand that a signature on a DID Document does NOT prove that it is (or ever was) the correct DID Document for a given DID. The only way to ensure that is to go through the DID Resolution process. See [Binding of Identity](https://w3c-ccg.github.io/did-spec/#binding-of-identity).

index.html
TomCJones: I believe that the above " first the signature MUST be valid before properties are returned" is not correct in the sense that signatures must have been valid at the time they were made, which doesn't require that they are currently valid. This mistake seems to be made over and over in ccg docs. It is important and must not be propagated. The CCG cannot override existing known good security practices.

AxelNennker: Could you please elaborate? If the resolver is tasked to return content of a signed document then that signature must be valid or the resolver must resolver must not return content of the document if the signature of the document is not valid.
If the signing key was revoked because it was compromised then the resolver should not return content but indicate an error this fatal condition.

What do you mean by the signature being once valid and not being valid at another point in time? Oscillating validity?

TomCJones: not at all - there are a variety of reasons for a signature to be invalid at a point in time. This biggest reason is that the validity period expired. If that is the reason, then the signature could have be valid at the time the documents was signed, but not at the current time. I would not accept the limitation that the did doc should fail to be returned in that case.

peacekeeper:
Just created issue https://github.com/w3c-ccg/did-resolution/issues/13 so we can also track this topic there.

jandrieu:
Closing. Thanks, Markus!

PR #69, opened 2018-03-13
Add note of owner field and underlying specifications. Fixes #25.

### LINKED ISSUE #25: Should/can we drop the owner property on embedded keys?
"Embedded" keys have an `owner` property that links back to their parent object in our examples, for example:

``` javascript
{
  "@context": "https;//w3id.org/did/v1",
  /* ... */
  "authenticationCredential": [{
    // this key can be used to authenticate as DID ...9938
    "id": "did:example:123456789abcdefghi#keys/1",
    "type": "RsaCryptographicKey",
  }]
  /* ... */
}
```

This is confusing some readers; why should an object (or rather its embedded object) link back to itself?

In Linked Data Signatures, when we sign/etc things we need to link to a specific key, but we also need to know what the owner of that key is, and for security reasons that requires both the owner and the key link to each other.  For example, of Manu makes a post saying "I like ravioli", the signature needs to link not just to Manu as the one signing it, which of Manu's keys specifically.  So, this makes sense in LDS land generally.

The confusion is, in the example of a DID Document such as above, we don't have a ravioli object pointing to a key pointing to its owner, we have a DID linking to (or appearing to "embed") a key and linking back to *itself*.

The "linking back to itself" is the bit that's causing great amounts of confusion. :)

We could maybe drop the requirement to link back to its owner assuming we're using an "embedded" fragment identifier... but this would be incompatible with Linked Data Signatures, which has good reason to have this for the general case.

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/69.html" title="Last updated on Apr 2, 2018, 8:35 PM GMT (c58df5c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/69/b0f2168...c58df5c.html" title="Last updated on Apr 2, 2018, 8:35 PM GMT (c58df5c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1082,7 +1082,7 @@ Example:
   }, {
     "id": "did:example:123456789abcdefghi#keys-2",
     "type": "Ed25519VerificationKey2018",
-    "owner": "did:example:123456789abcdefghi",
+    "owner": "did:example:pqrstuvwxyz0987654321",
     "publicKeyBase58": "H3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV"
   }, {
     "id": "did:example:123456789abcdefghi#keys-3",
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1074,6 +1094,14 @@ <h2>Public Keys</h2>
 }
 </pre>
 
+<p class="note">
+While the <code>owner</code> field may seem redundant in the examples above,
+keys may be expressed in a DID Document where the
+owner is described in another DID Document. Linked Data Proof libraries
+typically expect the <code>owner</code> field to always exist and may throw an
+exception if it is missing.
```
dlongley: +1 -- and we should add an example where the owner is NOT the DID document owner. We may also want to add some text:

"Futhermore, per the requirement that DID Documents be interpretable as either a graph or a tree, a default owner field cannot be inferred by using a key's position in a tree."

msporny: Added extra text and updated example. Done.

PR #74, opened 2018-04-30
Update RWoT leadership team

kimdhamilton:
Per recent discussions, we need to update the rebooting web of trust leadership team for current and future drafts. The key changes are:
1. Added Heather Vescent (checking with her that's the version of her name she wants to use) and myself
2. Remove people not currently part of leadership team. Tell me if I removed anyone that should remain; these are just the folks I haven't seen in rebooting over the past year.

Question: what is the ordering used here?

### DIFF
--- a/index.html
+++ b/index.html
@@ -135,8 +135,7 @@ endorsement should be inferred.
 <p>
 Work on this specification has also been supported by the Rebooting the
 Web of Trust community facilitated by Christopher Allen, Shannon Appelcline,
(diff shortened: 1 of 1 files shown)


vsnt:
+1 Kim. Heather Vescent is correct (and legal).

msporny:
@kimdhamilton are you okay w/ adding Kiara, Brian, and Betty back into facilitators... Also, we could do two different lists - one for leadership and another for facilitators. I think it's fine as is... thoughts?

kimdhamilton:
Definitely! This is great as you've done it, just wasn't sure how it worked. LGTM

PR #75, opened 2018-05-12
Propose service endpoint identification mechanism. Fixes #72.

### LINKED ISSUE #72: Remove 'name' from service description
The current consensus in the group seems to be that service descriptions looks something like this:

```
{
  "id": "did:v1:1234567",
  ... // DID Document goes here ...
  "service": [{
    "type": "urn:foobar:MessagingService",
    "name": "messaging",
    "serviceEndpoint": "https://example.com/messages/398tuo"
  }]
}
```

... which could result in the following URL that uses the service above:

```
did:v1:1234567;messaging/somepath#somefrag3
```
I suggest that we remove `name` and move to a less error prone format:

```
{
  "id": "did:v1:1234567",
  ... // DID Document goes here ...
  "service": [{
    "id": "did:v1:1234567;messaging",
    "type": "urn:foobar:MessagingService",
    "serviceEndpoint": "https://example.com/messages/398tuo"
  }]
}
```

Note that we get rid of the `name` field and use `id` instead. This approach has the following benefits:

1. It fits with the JSON-LD model more cleanly, 
2. doesn't require implementations to destructure a URL and look for specific properties (opaque identifiers), 
3. avoids the creation of a scenario where the `id` and the constructed service URL can differ, and 
4. enables us to use the semicolon syntax to address things that are not just service endpoints in the document.

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/75.html" title="Last updated on May 27, 2018, 5:51 PM GMT (e73f197)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/75/5f20c5d...e73f197.html" title="Last updated on May 27, 2018, 5:51 PM GMT (e73f197)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1239,7 +1239,7 @@ The value of the <code>service</code> property should be an array of service end
   </li>
 
   <li>
-Each service endpoint must include <code>id</code>, </code><code>type</code>, and <code>serviceEndpoint</code> properties, and MAY include additional properties.
+Each service endpoint must include <code>id</code>, <code>type</code>, and <code>serviceEndpoint</code> properties, and MAY include additional properties.
   </li>
 
   <li>
--- a/index.html
+++ b/index.html
@@ -1248,10 +1248,10 @@ specification.
   </li>
 
   <li>
-The value of the <code>serviceEndpoint</code> property MUST be a valid URI
-conforming to [[RFC3986]] and normalized according to the rules in section 6
-of [[RFC3986]] and to any normalization rules in its applicable URI scheme
-specification or a JSON-LD object.
+The value of the <code>serviceEndpoint</code> property MUST be a
+JSON-LD object or a valid URI conforming to [[RFC3986]] and normalized
+according to the rules in section 6 of [[RFC3986]] and to any normalization
+rules in its applicable URI scheme specification.
   </li>
 </ol>
 
--- a/index.html
+++ b/index.html
@@ -1239,7 +1239,7 @@ The value of the <code>service</code> property should be an array of service end
   </li>
 
   <li>
-Each service endpoint must include <code>type</code> and <code>serviceEndpoint</code> properties, and MAY include additional properties.
+Each service endpoint must include <code>id</code>, </code><code>type</code>, and <code>serviceEndpoint</code> properties, and MAY include additional properties.
   </li>
 
   <li>
@@ -1248,10 +1248,10 @@ specification.
   </li>
 
   <li>
-The value of the <code>serviceEndpoint</code> property MUST be a valid URI conforming to
-[[RFC3986]] and normalized according to the rules in section 6 of [[RFC3986]]
-and to any normalization rules in its applicable URI scheme specification or a
-JSON-LD object.
+The value of the <code>serviceEndpoint</code> property MUST be a valid URI
+conforming to [[RFC3986]] and normalized according to the rules in section 6
+of [[RFC3986]] and to any normalization rules in its applicable URI scheme
+specification or a JSON-LD object.
   </li>
 </ol>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1248,10 +1248,10 @@ <h2>Service Endpoints</h2>
...
 
   <li>
-The value of the <code>serviceEndpoint</code> property MUST be a valid URI conforming to
-[[RFC3986]] and normalized according to the rules in section 6 of [[RFC3986]]
-and to any normalization rules in its applicable URI scheme specification or a
-JSON-LD object.
+The value of the <code>serviceEndpoint</code> property MUST be a valid URI
+conforming to [[RFC3986]] and normalized according to the rules in section 6
+of [[RFC3986]] and to any normalization rules in its applicable URI scheme
+specification or a JSON-LD object.
```
kimdhamilton: nit: this is hard to parse. I'm interpreting it like the following; can we update this?

> The value of the <code>serviceEndpoint</code> property MUST be either (1) a valid URI conforming to [[RFC3986]] and normalized according to the rules in section 6 of [[RFC3986]] and to any normalization rules in its applicable URI scheme specification, or (2) a JSON-LD object.

Or possibly (assuming that interpretation is right) put (2) before (1), since (1) is quite long and causes the reader to lose context

msporny: Done.

... although even that language is confusing. This is a general cleanup that needs to happen as a part of the spec. The "rules" in the spec are overly prescriptive in some places while being wildly under-prescribed in other cases.

kimdhamilton:
need reviews from @peacekeeper and @talltree

index.html
```diff
@@ -1239,7 +1239,7 @@ <h2>Service Endpoints</h2>
   </li>
 
   <li>
-Each service endpoint must include <code>type</code> and <code>serviceEndpoint</code> properties, and MAY include additional properties.
+Each service endpoint must include <code>id</code>, </code><code>type</code>, and <code>serviceEndpoint</code> properties, and MAY include additional properties.
```
peacekeeper: Even if we decide to remove the `name` field and use `id` instead, I don't think `id` should be mandatory.

msporny:
@peacekeeper wrote:
> If you now invoke a resolver for DID did:example:123456789abcdefghi and ask for a service type MessagingService, would it return the above service? I would expect yes.

I would expect that the resolver spec would outline the algorithm that does the search, and yes, that algorithm would be a fairly easy one to write given that `MessagingService`s are easy to identify via the `type` property.

index.html
davidlehn: There's an extra `</code>` before `type`.

msporny: Fixed in https://github.com/w3c-ccg/did-spec/pull/75/commits/e73f197bc45e067af199136a91c9d486adfe51c2

msporny:
Pulling this in, it's been sitting out there for 14 days. We can improve/change if @talltree provides feedback at a later date.

index.html
```diff
@@ -1266,24 +1266,31 @@ <h2>Service Endpoints</h2>
 <pre class="example nohighlight" title="Various service endpoints">
 {
   "service": [{
+    "id": "did:example:123456789abcdefghi;openid",
```
Fak3: This is the first time i see semicolon in uri - is this some form of did path?

davidlehn: I think the intent is to be like a matrix paramter? It was mentioned in https://github.com/w3c-ccg/did-spec/issues/72 that a description of the `;` syntax is needed but that may have been forgotten.  I opened an issue related to using `key=value` matrix parameters vs just single keys: https://github.com/w3c-ccg/did-spec/issues/90.

PR #77, opened 2018-05-12
Added IETF registration of DID URI Scheme. Fixes #73.

### LINKED ISSUE #73: Register DID URI Scheme
We should register the DID URI scheme here:

https://www.iana.org/assignments/uri-schemes/uri-schemes.xml

msporny:
This is a complete registration template for IETF registration. I've already submitted it to IETF as a provisional registration, which means we can update it as the spec matures.

### DIFF
--- a/ietf-registration.txt
+++ b/ietf-registration.txt
@@ -1,0 +1,27 @@
+Scheme name: did
+Status: Provisional
+
+Applications/protocols that use this scheme name:
+
+A "did" URI is used to express an identifier that is conformant to the
+Decentralized Identifier[1] specification. The specification details how
+identifiers conformant to the "did" URI Scheme can be implemented using
+decentralized ledgers, decentralized hashtables, and other types of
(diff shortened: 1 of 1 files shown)


msporny:
I filled all those sections out and THEN saw the part that says that those sections are no longer required:

https://tools.ietf.org/html/rfc7595#section-7.4

> The previous version of this specification required the following additional fields in a scheme registration request.  These fields are no longer part of the template.  The answers instead belong in the scheme specification.

kimdhamilton:
sounds good

msporny:
The IETF Provisional URI Scheme Registration has been accepted: https://lists.w3.org/Archives/Public/public-credentials/2018May/0028.html

msporny:
Merging, because it's already been processed at IETF.

peacekeeper:
Is this an "IETF registration" or an "IANA registration"? Just confused by the file name.

PR #78, opened 2018-05-12
Add section on extensibility. Fixes #58.

### LINKED ISSUE #58: Explain the reasoning behind using graph data models and JSON-LD
From @rxgrant at RWoT6:

The spec currently doesn't explain why we're depending on graph-based data models and JSON-LD for extensibility. It also doesn't explain when you can and/or shouldn't process the document as pure JSON. We should add a section explaining this architectural decision. Mention the expectation that developers will parse DID Documents with common libraries (e.g. what happens when "owner" doesn't exist).

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/78.html" title="Last updated on May 15, 2018, 1:55 PM GMT (ecf73c6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/78/8f2bd19...ecf73c6.html" title="Last updated on May 15, 2018, 1:55 PM GMT (ecf73c6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -89,6 +89,12 @@
         inlineCSS: true
       };
     </script>
+    <style>
+pre .highlight {
+  font-weight: bold;
+  color: green;
+}
+    </style>
   </head>
   <body>
     <section id='abstract'>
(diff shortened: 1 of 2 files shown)


ChristopherA:
ACK, though I’d prefer a different example rather than favorite food as personal information should not be put into a DID Object. Instead something like a preferred contact address or something relevant to to a legacy identity (PGP fingerprint?). We don’t want to encourage DID Objects to become FOAF-like bags of personal information but instead have it point to the bags :-)

msporny:
> ACK, though I’d prefer a different example rather than favorite food as personal information should not be put into a DID Object. 

Hmm, so one of the first things I'm going to do is port my LinkedIn profile to a public DID Document... I think there are valid use cases.

That said, your point stands, we need a better example.

> Instead something like a preferred contact address or something relevant to to a legacy identity (PGP fingerprint?). 

... I'll try to brainstorm something that's not super geeky... maybe a public photo stream?

> We don’t want to encourage DID Objects to become FOAF-like bags of personal information but instead have it point to the bags :-)

I agree, with a few caveats. We really need a "best practices" guide.

msporny:
@ChristopherA wrote:
>  I’d prefer a different example

Done in https://github.com/w3c-ccg/did-spec/pull/78/commits/60782e3a3e757937d0bbd8dcba238ce7a243ed6b

ChristopherA:
@msporny commented:
>Hmm, so one of the first things I'm going to do is port my LinkedIn profile to a public DID Document... I think there are valid use cases.

In fact, with BTCR the common pattern is to point to a static file or IPFS object that is bag of self-signed verifiable credentials (which in fact may be appended to the DiD Object). But inside the BTCR DID Object only points points to it.

Longer term the DID Object pointer may include a clue to describe how to decrypt the file, such as using a mutual DH key. I really like the [miniLock](https://miniLock.io) [format](https://45678.github.io/miniLock-file-format/2.html] design, which I’d like to adapt to work with DID keys. This way you could only decrypt my Verifiable Credentials if you were in my web-of-trust set.

index.html
```diff
@@ -1454,6 +1460,165 @@ <h2>Proof (Optional)</h2>
...
+in a number of different ways:
+        </p>
+
+        <ul>
+          <li>
+The requirement to model complex multi-entity relationships is provided
+through the use of a graph-based data model.
+          </li>
+          <li>
+The requirement to be able to extend the machine-readable vocabularies
```
kimdhamilton: Minor readability suggestion; change this sentence to something like:

> The requirement to enable extending the machine-readable vocabularies used to describe information in the data model -- without relying on a centralized system -- is accomplished via the use of [LINKED-DATA].

index.html
```diff
@@ -1454,6 +1460,165 @@ <h2>Proof (Optional)</h2>
...
+          </li>
+          <li>
+The requirement to provide all of the extensibility mechanisms outlined above
+in a data format that is popular among software developers and web page authors
+is enabled via the use of [[!JSON-LD]].
+          </li>
+        </ul>
+
+        <p>
+This approach to data modelling is often called an "open world assumption",
```
kimdhamilton: just a time-saving note to other US reviewers who think "modelling" is a typo: it sounds like it's the preferred spelling in most parts of the world. :)

index.html
```diff
@@ -1454,6 +1460,165 @@ <h2>Proof (Optional)</h2>
...
+    "type": "PhotoStreamService",
+    "serviceEndpoint": "https://example.org/photos/379283"
+  }</span>]
+}
+        </pre>
+
+        <p>
+The examples so far have shown that it is easy to extend the Decentralized
+Identifiers Data Model in a permissionless and decentralized way. The
+mechanism shown also ensures that Decentralized Identifiers that were created
```
kimdhamilton: readability nit: can remove "shown" and "that were", which results in

>This mechanism also ensures that Decentralized Identifiers created in this way...

index.html
```diff
@@ -1454,6 +1460,165 @@ <h2>Proof (Optional)</h2>
...
+semantic ambiguity.
+        </p>
+
+        <p>
+An extensibility model that is this dynamic does increase implementation burden.
+Software written for such a system will have to determine if accepting
+<a>DID Document</a>s with extensions is acceptable based on the
+risk profile of the application. Some applications may choose to accept but
+ignore extensions, others may choose to only accept certain extensions, while
+highly secure environments may require that no extensions are allowed. These
```
kimdhamilton: readability nit: 

Change:
> highly secure environments may require that no extensions are allowed. 

to:
> highly secure environments may disallow extensions.

index.html
```diff
@@ -1454,6 +1460,165 @@ <h2>Proof (Optional)</h2>
...
+decisions are up to the developers of these applications and are specifically
```
kimdhamilton: nit: 
Change:
> are up to the developers of these applications

To:
> are up to the application developers

index.html
msporny: @msporny thanks his British colonial upbringing in Sri Lanka for getting that right.

index.html
msporny: Done. https://github.com/w3c-ccg/did-spec/pull/78/commits/ecf73c6addd36d2f300a4f74992e55919a371b7e

index.html
msporny: Also removed "provide a mechanism to".

Fixed in https://github.com/w3c-ccg/did-spec/pull/78/commits/ecf73c6addd36d2f300a4f74992e55919a371b7e.

index.html
msporny: Fixed in https://github.com/w3c-ccg/did-spec/pull/78/commits/ecf73c6addd36d2f300a4f74992e55919a371b7e.

index.html
msporny: Fixed in https://github.com/w3c-ccg/did-spec/pull/78/commits/ecf73c6addd36d2f300a4f74992e55919a371b7e.

Thanks for all the editorial suggestions, @kimdhamilton!

msporny:
Merging, we can modify the section at a later date if @rxgrant wants further changes.

PR #79, opened 2018-05-12
Add key search algorithm as requested by @mikelodder7. Fixes #51.

### LINKED ISSUE #51: authorization section with publicKey
I'm wondering if publicKey in the authorization section should be renamed to publicKeyURI to make it more clear about what it is. 
Is there a use case where the actual key would be put there instead of the publicKey section?
All of our discussions so far have used it as a DID url for the key either in the same DID document or somewhere else. To keep things clean I would recommend that keys go in the publicKey section and authorization just uses DID urls to refer to them.
I'm not happy about publicKeyURI but I believe the name publicKey is not a good descriptor.

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/79.html" title="Last updated on May 15, 2018, 2:08 PM GMT (34dd3ba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/79/8f2bd19...34dd3ba.html" title="Last updated on May 15, 2018, 2:08 PM GMT (34dd3ba)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1152,7 +1152,8 @@ If the <code>id</code> property of the object matches <em>value</em>, set
   <li>
 If <em>result</em> does not contain at least the <code>id</code>,
 <code>type</code>, and <code>owner</code> properties as well as any
-mandatory public cryptographic material, throw an error.
+mandatory public cryptographic material, as determined by the
+<em>result</em>'s <code>type</code> property, throw an error.
   </li>
 </ol>
 
--- a/index.html
+++ b/index.html
@@ -1088,6 +1088,74 @@ Example:
 }
 </pre>
 
+<p>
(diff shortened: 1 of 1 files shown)


kimdhamilton:
Possibly this is obvious to readers, but small suggestion for clarity:
Instead of:
> as well as any mandatory public cryptographic material,

Change to something like:
> as well as any mandatory public cryptographic material, as determined by the publicKey `type`,

msporny:
@kimdhamilton wrote:
> Change to something like

Fixed in https://github.com/w3c-ccg/did-spec/pull/79/commits/34dd3baf9d548606e91013de35ef043bcb832edf.

msporny:
Thanks @mikelodder7 -- merged.

Issue #90, opened 2018-06-13
Service endpoint identification matrix parameter


davidlehn:
The endpoint identification format of `did:example:123456789abcdefghi;openid` was introduced in https://github.com/w3c-ccg/did-spec/pull/75.  The `;` usage may cause URI parsers to expect matrix parameters. I think matrix parameters are still a proposal [1] vs an actual spec. It does look like the somewhat rare usage in the wild uses `key=value` pairs [2][3][... many search results ...].  In the 1996 proposal there is an example that considers a relative matrix param that's just a key as something that blanks out that key. It may make sense to use `key=value` syntax in dids to align with other usage. Could be as simple as `service=openid`.

https://github.com/w3c-ccg/did-spec/issues/72 mentions the need to describe the `;` syntax, but that has not been addressed.

[1] https://www.w3.org/DesignIssues/MatrixURIs.html
[2] https://stackoverflow.com/questions/401981/when-to-use-query-parameters-versus-matrix-parameters
[3] https://github.com/medialize/URI.js/issues/181

Fak3:
Couldn't find a rationale for the need of semicolon syntax in #72. Can the `;openid` part be replaced with just fragment: `#service-openid` ?

msporny:
> Could be as simple as service=openid

This is what I had counter-proposed, but folks pushed back for reasons I can't remember anymore. I think the push back was that it would be impossible to tell when the DID-based URL ends and the HTTP-based URL begins.

@talltree @mikelodder7 @peacekeeper -- any thoughts on this?

msporny:
> Can the ;openid part be replaced with just fragment: #service-openid

In that scenario, how do you know when the DID-based URL ends and the HTTP-based URL begins. For example:

```
did:example:234786;vcrepository=/foo/bar/public-credentials.jsonld
```

Would you do something like this?

```
did:example:234786?service=vcrepository&path=/foo/bar/public-credentials.jsonld
```

Fak3:
> In that scenario, how do you know when the DID-based URL ends and the HTTP-based URL begins.

Why someone needs to construct URL like this? Embedding one URL into another

Fak3:
isn't "serviceEndpoint" property supposed to carry the url of the service? Why put it in the "@id"?

dmitrizagidulin:
Have we come to any consensus on this issue? (Would be helpful to have this resolved, so we can validate service endpoint IDs in our code.)

mikelodder7:
[...]

PR #95, opened 2018-08-02
Added section on did service component

mikelodder7:
Signed-off-by: Michael Lodder <redmike7@gmail.com>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/95.html" title="Last updated on Aug 2, 2018, 8:24 PM GMT (272ab00)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/95/6c50ae4...272ab00.html" title="Last updated on Aug 2, 2018, 8:24 PM GMT (272ab00)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -101,6 +101,12 @@
         url: "https://www.linkedin.com/in/markus-sabadello-353a0821",
         company: "Danube Tech",
         companyURL: "https://danubetech.com/"
+      },
+      {
+        name: "Michael Lodder",
+        url: "www.linkedin.com/in/mike-lodder-48aab67",
+        company: "Evernym",
+        companyURL: "https://evernym.com/"
       }
     ],
 
@@ -127,6 +133,9 @@
   pre .highlight {
     font-weight: bold;
     color: green;
+  }
+  .noun {
+    font-family: 'Courier New';
   }
   </style>
 </head>
@@ -1439,7 +1448,6 @@ any normalization rules in its applicable URI scheme specification.
       <p>
 Example:
       </p>
-
       <pre class="example nohighlight" title="Various service endpoints">
 {
   "service": [{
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -128,6 +134,9 @@
     font-weight: bold;
     color: green;
   }
+  .noun {
```
msporny: We try very hard not to create new CSS classes and reuse the ones provided by ReSpec. I suggest striking this CSS class and using "&lt;code>" instead.

index.html
```diff
@@ -1487,8 +1495,65 @@ <h2>
...
 considerations regarding authentication service endpoints.
       </p>
-    </section>
+      <p>
+The purpose of the <span class="noun">did-service</span> component may optionally follow a DID is to enable
+construction of a DID reference that may be algorithmically transformed into a
+<em>concrete URL</em> to access the target resource. There are two algorithms for this transformation.
+Both begin with the same first step:
+      </p>
+<ol>
```
msporny: Algorithms in W3C specs tend to be written using a common language (which is unfortunately tribal knowledge, for the most part). 

Here's an example of one that's been through the WG process: https://www.w3.org/TR/json-ld-api/#expansion-algorithm

... and another one in the HTML5 spec on how to parse URLs:  https://www.w3.org/TR/html52/infrastructure.html#parsing-urls

I suggest restructuring the algorithms you wrote in the manner above as it'll be easier to read, understand, and get through the WG process.

index.html
```diff
@@ -1487,8 +1495,65 @@ <h2>
...
+  <li>
+  Extract the DID plus the the <span class="noun">did-service</span> component from the DID reference.
```
msporny: How do you extract it? Split it on the first semicolon? Split on colons and then the third element, and then a semicolon? W3C specs are painfully precise... if you're not precise, the likelihood that someone will misimplement something and deploy it to millions of people is very high (and then the entire community has to live with the side effects of that terrible implementation forever).

index.html
```diff
@@ -1487,8 +1495,65 @@ <h2>
...
+  Call this the <em>DID service locator</em>. Call the rest of the DID reference the <em>service-specific string.</em>
+  </li>
+</ol>
+<h3>Service Selection by ID</h3>
+<p>
+  This algorithm MUST be used if the <span class="noun">did-service</span> component does NOT begin with the string <span class="noun">type=</span>
+</p>
+<ol>
+  <li>Select the first <code>service</code> object whose id property contains an exact match to the DID service locator.</li>
```
msporny: This is a graph model, so you can't say "first" as the list of services is a set (not an ordered array). You could select based on lexicographical order. Also note that objects with the same id property get merged in a graph-based data model. I suggest you just say something to the effect of "Select the service object whose id property is an exact match for..."

index.html
```diff
@@ -1487,8 +1495,65 @@ <h2>
...
+  did:example:123456789abcdefghi;agent/some/path?query#frag
+</pre>
+<p>A DID resolver would algorithmically transform that DID reference to the following concrete URL:</p>
+<pre class="example nohighlight">
+  https://agent.example.com/8377464/some/path?query#frag
+</pre>
+<h3>Service Selection by Type</h3>
+<p>This algorithm MUST be used if the <span class="noun">did-service</span> component begins with the string <span class="noun">type=.</span></p>
+<ol>
+  <li>Select the first <code>service</code> object whose type property contains an exact match
```
msporny: service "array" is a set, not a list, so selecting the first item isn't necessary.

index.html
```diff
@@ -1487,8 +1495,65 @@ <h2>
...
+      There MAY be more than one <code>service</code> object that meets this criteria––see below.</li>
+  <li>Select the <code>serviceEndpoint</code> property of the selected <code>service</code> object.</li>
+  <li>Extract the value of the <code>serviceEndpoint</code> property. Call this the endpoint URL.</li>
+  <li>Append the service-specific string to the endpoint URL.</li>
+  <li>The final result is the concrete URL.</li>
+</ol>
+<p>
+  If there is more than one <code>service</code> object that meets the selection criteria in step 2 above,
+  a DID resolver SHOULD return an array of concrete URLs. It is RECOMMENDED that service endpoint
+  protocols apply index-based priority to this array, i.e., the highest priority concrete URL is the lowest position in the array.
```
msporny: You can't deterministically do index-based priority on graph-based data models containing sets of information. I suggest you do lexicographical sorting... but even that seems problematic (and doesn't make sense for id-based selections). We may need to requires that resolutions that result in multiple values require an external selection input into the algorithm.

talltree:
Manu, all your feedback seems reasonable. I will work with @mikelodder7 and @dhh1128 to suggest fixes. RE the use case for "type=", the source is Daniel Buchner. He is adamant that it is need to select service endpoints for the Hub spec being developed at DIF.

peacekeeper:
As a meta-point, @dmitrizagidulin and I started a separate DID Resolution doc, since that's a pretty big topic by itself: https://w3c-ccg.github.io/did-resolution/. I think algorithms for DID resolution, service selection, target URL construction, etc. should go in there rather than in the main DID spec doc. I propose that we:
1. First make sure there's a conclusion on the syntax for `did-service` (see the ongoing discussion in issue #90).
2. Reduce this PR to only describe the syntax.
3. Describe DID Resolution logic in the separate doc, I'd be happy to work with @dmitrizagidulin, @mikelodder7 on a first draft.

rhiaro:
I can't see anything in this PR that isn't either going to be moved to the DID Resolution spec or will be addressed with a new PR when #90 is resolved, so I'm closing this.

Issue #101, opened 2018-09-14
Remove "identity owner" as a term in the spec


jandrieu:
The term "identity owner" is problematic in a few ways.

The simplest, practical argument is that proof of control (via private key) does not actually establish or prove ownership. Anyone could easily steal a private key and assume control over a DID. That would not make them the owner of the DID, which is a legal question and therefore a matter for the courts and evidentiary proceedings.

My second reason for removing "identity owner" is that it violates the functional notions of identity. Identity is how we recognize, remember, and respond to specific people and things. The only owner of that is the *observer*, the person or system doing the recognizing, remembering, and responding. This is 180 degrees out of phase with the wishful thinking that the self-sovereign individual "owns" their identity. Don't get me wrong, I'm a big fan of self-sovereign identity; it just has nothing to do with ownership.

The third reason is that it violates the subjective notion of identity, which recognizes that identity is literally manifest solely at at the point of identification. It isn't a thing outside that recognition. In contrast, the reification of identity as a concrete thing, such as might be represented digitally or in a physical credential is, in mathematical terms, a projection from a higher dimensional notion of identity to at best a "digital identity" or, depending on the situation, an identifier, a credential, a token, etc. Referring to these lesser dimensional projections of identity *as* identity consistently confuses people. Identifiers are not identities. Credentials are not identities. Tokens are not identities. Yes, many many people in our industry use that shorthand because it is convenient when talking with others "in the know". But it lacks rigor and experience has shown that such use is a huge barrier to understanding for laypeople.

The final reason is that even if the above points don't resonate with you, they do illustrate a highly debatable and controversial point.

We only use the term 3 times in the spec. I have no doubt we can replace those three phrases with something clearer and more accurate.

peacekeeper:
I agree the term "identity owner" can be problematic, because of the cases you mention where control != ownership.

But I'm not sure I fully agree with the functional/subjective identity concept that always depends on the observer. How would DID Auth fit into this model? In the case of [DID Auth and the little I-am-me](https://medium.com/@markus.sabadello/did-auth-and-the-little-i-am-me-ec14d757ff09), none of the other animals was able to issue a verifiable credential, but still it has its own independent existence! Sorry this may be a bit off-topic. Perhaps let's have a discussion about this at IIW :)

jandrieu:
[...]

PR #102, opened 2018-09-15
Replace "owner" with "controller"

jandrieu:
In addressing #101, it was clear that ownership and owner were used throughout despite the problems cited in the issue: namely that proof of control does not establish ownership, just control.

This PR adjusts the language from owner to controller for better accuracy and understandability, both in the spec-text and in the field name.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/102.html" title="Last updated on Sep 15, 2018, 1:23 AM GMT (60ce842)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/102/4b0d5e9...60ce842.html" title="Last updated on Sep 15, 2018, 1:23 AM GMT (60ce842)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -283,7 +283,7 @@ de-referenced to locate a representation of a resource on the Web
 [[URN]] is the term for a specific type of URI intended to
 persistently identify a resource, i.e., an identifier that will
 never change no matter how often the resource moves, changes labels,
-changes owners, etc. URNs are intended to last forever.
+changes controllers, etc. URNs are intended to last forever.
           </li>
         </ol>
       </section>
@@ -311,7 +311,7 @@ such authority.
           </li>
 
           <li>
-A URL whose ownership and associated metadata, including public
+A URL whose control and associated metadata, including public
 keys, can be cryptographically verified. Authentication via DIDs
 and DID Documents leverage the same public/private key cryptography
 as distributed ledgers.
@@ -339,7 +339,7 @@ two".
 There are of course many use cases where it is desirable to
 discover a DID when starting from a human-friendly identifier—a
 natural language name, a domain name, or a conventional address for
-a DID owner such as a mobile telephone number, email address,
+a DID controller such as a mobile telephone number, email address,
 Twitter handle, or blog URL. However, the problem of mapping
 human-friendly identifiers to DIDs (and doing so in a way that can
 be verified and trusted) is out-of-scope for this specification.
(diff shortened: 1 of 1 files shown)


msporny:
@jandrieu wrote:
> Although it is a fine foundation for our work, I have a hard time with an argument that a current draft of a non-standards track spec should be read in as "required" and immutable.

You're missing @dlongley's point. He's saying that Mastodon shipped Linked Data Signatures (the current wobbly spec) to 1 million users. So, whether we like it or not, that version of the spec exists now and we are busy supporting it. The change you're suggesting, which I'm taking no position on currently, would require Digital Bazaar to support two different code paths which may result in security attacks that we haven't considered yet.

One of his points is noting the cost (to implementers) of switching 1 million users that are using already deployed software over to what you're suggesting.

I think you're making some good points in this thread... just make sure you're understanding @dlongley's points fully as well.

jandrieu:
@msporny Fair enough. I recognize there may be some non-trivial switching costs. IMO, better to incur that now before we deploy to another million users, or 10 million, or more.

ChristopherA:
I strongly lean to getting this right. I agree with Joe that this is an
ongoing problem and that it is time to fix it.

— Christopher Allen [via iPhone]

jandrieu:
@dlongley Another reference to the complications of using "owner" in reference to data, especially in the context of identity, recently popped up on the VRM list. That thread pointed to a current article by Elizabeth Renieris and Dazza Greenwood.  Worth reading.
https://medium.com/@hackylawyER/do-we-really-want-to-sell-ourselves-the-risks-of-a-property-law-paradigm-for-data-ownership-b217e42edffa

OR13:
Seems like controller won out... https://github.com/digitalbazaar/jsonld-signatures/pull/45/files

Can we adopt this change? or make it clear which version of jsonld / jsonld signatures this spec is based on that uses the term controller?

dlongley:
> Seems like controller won out... https://github.com/digitalbazaar/jsonld-signatures/pull/45/files
> 
> Can we adopt this change? or make it clear which version of jsonld / jsonld signatures this spec is based on that uses the term controller?

I'm +1 to adopting "controller" now -- and as you can see above our implementation of LD signatures/proofs has adopted that term and added backwards compatibility support for "creator". There are a number of other terms related to this that will be discussed on an upcoming CCG call.

Related to https://github.com/w3c-ccg/community/issues/50.

mwherman2000:
When we consider Things (e.g. cars, pets, houses, business documents, products, assemblies, parts, etc.) in addition to Actors (e.g. Persons, Organizations, Software Agents), how does this influence the choice of Owner vs. Controller? ...Things and ownership is an important scenario that is not addressed well in the current draft DID spec and the current topic is very relevant to rectifying this situation.

image

dmitrizagidulin:
(👍 from me on this PR)

rhiaro:
Merging; the group has consensus around this change.

Issue #85, opened 2018-05-23
Did Query


SmithSamuelM:
This action item from the Did Spec Completion proposals doc did not make it into the latest did spec 0.1

#2: DID URI Query Component
Motivation
We forgot to include the URI query component in defining DID ABNF. There was no reason for this; it simply an oversight.
Proposal
Add the DID query component to the ABNF in section 3.1, and add a section to define it.
  did-reference      = did [ ";" did-service ] [ "/" did-path ] [ "?" did-query ] 
                     [ "#" did-frag ]
  did                = "did:" method ":" specific-idstring
  method             = 1*namechar
  namechar           = %x61-7A / DIGIT
  specific-idstring  = idstring *( ":" idstring )
  idstring           = 1*idchar
  idchar             = ALPHA / DIGIT / "." / "-"
  did-service        = 1*servchar *( ";" 1*servchar )
  servchar           = idchar / "=" / "&"

SmithSamuelM:
The presence of the DID query has two main syntactical interpretations. 

When a DID path component is present the DID query is a modifier on the did path resource in the same way that it works on a URL.

When a DID path component is not present then the DID Query is a modifier on the Did meta-data obtained from the DDO and narrowed by the did fragment if present.

SmithSamuelM:
The query string can be used for any query purpose after the id-string, the md-string, or the path-segment.

SmithSamuelM:
What is remaining to close and merge this pull request?  The Query was always meant to be part of the DID syntax but was left out as an oversight.

peacekeeper:
I think [PR 106](https://github.com/w3c-ccg/did-spec/pull/106) nicely fixes the ABNF, but..
1. There are a number of other small issues with the PR which I have articulated there (e.g. in a few places it seems to assume that fragment comes before query?).
2. While adding `did-query` to the ABNF seems uncontroversial, there is more discussion about `did-service` [elsewhere](https://github.com/w3c-ccg/did-spec/issues/90).

Perhaps @mikelodder7 can update his PR accordingly, or we start a new PR.

mikelodder7:
Okay, I'll do that and try to get to pushing an update today.

mwherman2000:
[...]

PR #106, opened 2018-10-10
Add did-query (addresses #85)

mikelodder7:
Signed-off-by: Michael Lodder <redmike7@gmail.com>

Addresses issue [85](https://github.com/w3c-ccg/did-spec/issues/85)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/106.html" title="Last updated on Feb 5, 2019, 2:28 PM UTC (de0bcf8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/106/6c50ae4...de0bcf8.html" title="Last updated on Feb 5, 2019, 2:28 PM UTC (de0bcf8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -719,16 +719,18 @@ Following is the ABNF definition using the syntax in [[RFC5234]]
       </p>
 
       <pre class="nohighlight">
-did-reference      = did [ "/" did-path ] [ "#" did-fragment ]
+did-reference      = did [";" did-service ] [ "/" did-path ] ["?" did-query ] [ "#" did-fragment ]
 did                = "did:" method ":" specific-idstring
 method             = 1*methodchar
 methodchar         = %x61-7A / DIGIT
 specific-idstring  = idstring *( ":" idstring )
 idstring           = 1*idchar
 idchar             = ALPHA / DIGIT / "." / "-"
+did-service        = 1*servicechar *(";" 1*servicechar)
+servicechar        = idchar / "=" / "&amp;"
 </pre>
       <p>
-See Sections <a href="#paths"></a> and <a href="#fragments"></a> for the ABNF rules defining DID paths and
+See Sections <a href="#paths"></a>, <a href="#fragments"></a>, <a href="#query"></a> for the ABNF rules defining DID paths and
 fragments.
       </p>
     </section>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -797,6 +799,26 @@ <h2>
...
+    <h2>
+Query
+    </h2>
+A generic <a>DID query</a> (the did-query rule in Section <a href="#the-generic-did-scheme"></a>) is identical to a URI search and MUST
+conform to the ABNF of the query rule in [[RFC3986]].
+<p>If a DID is followed immediately by a DID query component
+(i.e., it does NOT contain a path or fragment),
+then the DID query operates directly against the DID document.</p>
+
+<p>If a DID is followed by a DID fragment followed immediately by a DID query component,
```
peacekeeper: This paragraph doesn't make sense to me. `did-fragment` comes after `did-query`.

index.html
```diff
@@ -797,6 +799,26 @@ <h2>
...
+then the DID query operates against the branch of the DID document
+specified DID document fragment.</p>
+
+<p>A DID method specification MAY defined its own DID fragment query syntax.</p>
```
peacekeeper: What's the motivation here?

index.html
```diff
@@ -797,6 +799,26 @@ <h2>
...
     </section>
 
+    <section>
+    <h2>
+Query
+    </h2>
+A generic <a>DID query</a> (the did-query rule in Section <a href="#the-generic-did-scheme"></a>) is identical to a URI search and MUST
+conform to the ABNF of the query rule in [[RFC3986]].
+<p>If a DID is followed immediately by a DID query component
+(i.e., it does NOT contain a path or fragment),
```
peacekeeper:
```suggestion
(i.e., it does NOT contain a path),
```

davidlehn:
Would be good to rewrite the commit message to say what patch does vs just saying it addresses an issue number.

dmitrizagidulin:
Addressed/refactored in PR #168.

index.html
SmithSamuelM: I think the point is that the query operations on what the path resolves too. Currently if there is no path the default is the path resolves to the DDO likewise if there is not path then the fragment resolves against the DDO.  The query applies to what ever the path resolves too. The query usually does not apply to a fragment but to the resource.  However because the did fragment has special semantics we could give special  meaning to the query if a did fragment is present.  I like the change does not contain a path and let the implementer decide what to do if a did fragment is present but not path.

peacekeeper:
Replaced by PR https://github.com/w3c-ccg/did-spec/pull/189.

Issue #121, opened 2018-12-20
[PURPOSE OF THE SPECIFICATION] Is this draft specification trying to address too many topics when there should be more than 1 spec


mwherman2000:
In https://w3c-ccg.github.io/did-spec/#purpose-of-this-specification,  it states...

> The first purpose of this specification is to define the generic DID scheme and a generic set of operations on DIDs that can be implemented for any distributed ledger or network capable of supporting DIDs. The second purpose of this specification is to define the conformance requirements for a DID method specification—a separate specification that defines a specific DID scheme and specific set of DID record operations for a specific distributed ledger or network.

There are multiple separable purposes/goals/tasks here (IMHO)...

1. Specification for a (basic) DID as a string-based identifier: `did:xyz:1234567890` - i.e. the "did:" identifier scheme spec
2. Specification of an "access protocol" for working with DID Documents (similar to the HTTP or SMTP specification).  This specification can and must reference #1.  But they need to be very clearly separate from each other.  I go as far as advocating for a second scheme to highlight that this is for a related but completely different purpose: i.e. the "didp:" or "drap:" schema/protocol (decentralized resource access protocol).  This spec/protocol is about how to _register, resolve, update, or revoke a_ *DID Entity/DID Document* (the latter being the JSON-LD serialization of the former) ...and *not* about _how to register, resolve, update, or revoke the identifier_.

That's it for now.  Perhaps these two separable purposes/goals/tasks can be covered in a single document but there are in fact 2 different protocols.  Whether this can be address by a single specification document is most likely a W3C governance/policy question I think.

Reference: [Hyperledger Indy/Sovrin Comprehensive Architecture Reference Model (INDY ARM) - latest version](https://hyperonomy.com/2019/01/13/hyperledger-indy-sovrin-comprehensive-architecture-reference-model-arm/) - bullets (12) thru (16) in both the diagram, Narration, and principles.

mwherman2000:
A key advantage of have two separate (logical and/or physical) specifications is that each spec (the identifier spec and the access protocol spec) can evolved on different time frames/parallel paths.

mwherman2000:
For example, the format of an email message (e.g. [RFC5322](https://tools.ietf.org/html/rfc5322)) is separate/separated from the SMTP protocol (e.g. [RFC5321](https://tools.ietf.org/html/rfc5321)) that operates on email messages. 

[...]

Issue #97, opened 2018-08-06
need rationale in spec for fully qualified DID references as the value of "id" fields


dhh1128:
After a reading of the spec, I am somewhat mystified by the usage of fully qualified DID references as the value of an "id" field. We have stuff like this from example 8:

```
 "service": [{
    "id": "did:example:123456789abcdefghi;openid",
 ``` 

... instead of:

```
 "service": [{
    "id": "openid",
 ``` 

... and stuff like this from example 5:

```
  "publicKey": [... {
    "id": "did:example:123456789abcdefghi#keys-2",
    "type": "Ed25519VerificationKey2018",
    "owner": "did:example:pqrstuvwxyz0987654321",
    "publicKeyBase58": "H3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV"
  }, {
```

When I would have expected:

```
  "publicKey": [... {
    "id": "keys-2",
    "type": "Ed25519VerificationKey2018",
    "owner": "did:example:pqrstuvwxyz0987654321",
    "publicKeyBase58": "H3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV"
  }, {
```

This second chunk of code is particularly interesting as an example because it references an owner that is NOT the owner of the provided DID Doc. The normative URI for such a key is that other owner's DID doc (perhaps at `did:example:pqrstuvwxyz0987654321#keys-407`). The fact that we allow a fully qualified DID reference here makes me think this ought to be possible--but apparently we have a constraint that every value of "id" inside the doc must refer to an item inside the doc, despite being fully qualified?

It feels like we are confusing the old SGML/XML/HTML notions of ID and IDREF; we're using an IDREF where we should be using an ID. In html, the value of a link take is an ID, and the hyperlink that points to it is an IDREF. You'd never look in an html doc and see `<a name="http://mydomain.com/path/to/my/doc#fragment"...>`; you'd just expect to see `<a name="fragment">`.

Now, perhaps this analysis is missing something and there is a compelling reason why we need fully qualified values. If that is the case, then I think the spec needs a paragraph or two explaining why. My major point is that such a pattern is surprising enough that it needs to be justified and documented. My minor point is that it may be the wrong decision in the first place.

msporny:
> It feels like we are confusing the old SGML/XML/HTML notions of ID and IDREF; we're using an IDREF where we should be using an ID.

[...]

PR #159, opened 2019-01-24
Updates to URIs, URLs, and URNs section

### LINKED ISSUE #155: [URIs, URLs, and URNs] Section doesn't conclude with a precise declaration of what a DID is.
In https://w3c-ccg.github.io/did-spec/#uris-urls-and-urns, the section doesn't conclude with a precise declaration of what a DID is.

My understanding is that the _id (DID)_ attribute (aka "DID") of a DID Document/DID Entity is actually a URN ...a subclass of URI ...and specifically not a URL.  URL is a centralized concept.   A URL is a locator.

### LINKED ISSUE #156: [URIs, URLs, and URNs] "DIDs have a foundation in URLs" is confusing because DIDs aren't URLs
In https://w3c-ccg.github.io/did-spec/#uris-urls-and-urns, it states...

> DIDs have a foundation in URLs, so it's important to understand...

This is interesting history but the wording of this sentenceis  confuding.  It leads the reader to believe that DIDs are URLs. See https://github.com/w3c-ccg/did-spec/issues/155

mwherman2000:
Fixes #155 
Fixes #156

***
<a href="https://pr-preview.s3.amazonaws.com/mwherman2000/did-spec/pull/159.html" title="Last updated on Jan 25, 2019, 4:30 PM UTC (9ab5825)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/159/dfdac17...mwherman2000:9ab5825.html" title="Last updated on Jan 25, 2019, 4:30 PM UTC (9ab5825)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -259,8 +259,9 @@ URIs, URLs, and URNs
         </h2>
 
         <p>
-DIDs have a foundation in URLs, so it's important to understand how
-the W3C <a href="https://www.w3.org/TR/uri-clarification/">clarified</a> the terms
+Given the potentially confusing relationship between DIDs, URIs, URLs and URNs, 
+so it's important to understand how the
+W3C <a href="https://www.w3.org/TR/uri-clarification/">clarified</a> the terms
 [[URI]] (Uniform Resource Identifier), [[URL]] (Uniform Resource
 Locator), and [[URN]] (Uniform Resource Name) in September 2001.
 The key difference between these three categories of identifiers
@@ -286,6 +287,11 @@ never change no matter how often the resource moves, changes labels,
 changes controllers, etc. URNs are intended to last forever.
           </li>
         </ol>
+        <p>
+DIDs are valid URIs; more specifically, they are also URLs and URNs. 
+DIDs are URLs because they are resolvable (dereferenceable).
+DIDs are also URNs because they are persistent identifiers. 
+        </p>
       </section>
 
       <section>
--- a/index.html
+++ b/index.html
@@ -288,7 +288,7 @@ changes controllers, etc. URNs are intended to last forever.
           </li>
         </ol>
         <p>
-DIDs are valid URIs; more specifically, they are also URLs and URNs. 
+DIDs are valid URIs; more specifically, they are also valid URLs and URNs. 
 DIDs are URLs because they are resolvable (dereferenceable).
 DIDs are also URNs because they are persistent identifiers. 
         </p>
--- a/index.html
+++ b/index.html
@@ -260,7 +260,7 @@ URIs, URLs, and URNs
 
         <p>
 Given the potentially confusing relationship between DIDs, URIs, URLs and URNs, 
-so it's important to understand how the
+it's important to understand how the
 W3C <a href="https://www.w3.org/TR/uri-clarification/">clarified</a> the terms
 [[URI]] (Uniform Resource Identifier), [[URL]] (Uniform Resource
 Locator), and [[URN]] (Uniform Resource Name) in September 2001.
--- a/index.html
+++ b/index.html
@@ -259,11 +259,10 @@ URIs, URLs, and URNs
         </h2>
 
         <p>
-Given the potentially confusing relationship between DIDs, URIs, URLs and URNs, 
-it's important to understand how the
-W3C <a href="https://www.w3.org/TR/uri-clarification/">clarified</a> the terms
+It is important to understand the relationships between DIDs, URIs, URLs and URNs. 
+In September 2001, the W3C <a href="https://www.w3.org/TR/uri-clarification/">clarified</a> the terms
 [[URI]] (Uniform Resource Identifier), [[URL]] (Uniform Resource
-Locator), and [[URN]] (Uniform Resource Name) in September 2001.
+Locator), and [[URN]] (Uniform Resource Name).
 The key difference between these three categories of identifiers
 are:
         </p>
@@ -289,8 +288,7 @@ changes controllers, etc. URNs are intended to last forever.
         </ol>
         <p>
 DIDs are valid URIs; more specifically, they are also valid URLs and URNs. 
-DIDs are URLs because they are resolvable (dereferenceable).
-DIDs are also URNs because they are persistent identifiers. 
+DIDs are URLs because they are resolvable (dereferenceable); they are also URNs because they are persistent identifiers. 
         </p>
       </section>
 
--- a/index.html
+++ b/index.html
@@ -263,7 +263,7 @@ It is important to understand the relationships between DIDs, URIs, URLs and URN
 In September 2001, the W3C <a href="https://www.w3.org/TR/uri-clarification/">clarified</a> the terms
 [[URI]] (Uniform Resource Identifier), [[URL]] (Uniform Resource
 Locator), and [[URN]] (Uniform Resource Name).
-The key difference between these three categories of identifiers
+The key differences between these three categories of identifiers
 are:
         </p>
 
--- a/terms.html
+++ b/terms.html
@@ -27,7 +27,8 @@ not require a centralized registration authority because it is
 registered with <a>distributed ledger</a> technology or other form of
 decentralized network. The generic format of a DID is defined in this
 specification. A specific <a>DID scheme</a> is defined in a
-<a>DID method</a> specification.
+<a>DID method</a> specification. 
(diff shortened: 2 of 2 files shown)


index.html
```diff
@@ -259,8 +259,9 @@ <h2>
         </h2>
 
         <p>
-DIDs have a foundation in URLs, so it's important to understand how
-the W3C <a href="https://www.w3.org/TR/uri-clarification/">clarified</a> the terms
+Given the potentially confusing relationship between DIDs, URIs, URLs and URNs, 
+so it's important to understand how the
```
davidlehn:
```suggestion
it's important to understand how the
```

index.html
```diff
@@ -286,6 +287,11 @@ <h2>
 changes controllers, etc. URNs are intended to last forever.
           </li>
         </ol>
+        <p>
```
msporny: Editorial nit: Would prefer to see this in a bulleted list of some kind since "DIDs are" is repeated unnecessarily.

index.html
```diff
@@ -259,8 +259,9 @@ <h2>
         </h2>
 
         <p>
-DIDs have a foundation in URLs, so it's important to understand how
-the W3C <a href="https://www.w3.org/TR/uri-clarification/">clarified</a> the terms
+Given the potentially confusing relationship between DIDs, URIs, URLs and URNs,
```
msporny: Would rather we not start with a negative spin here.

mwherman2000:
> Fine w/ it being merged, note the editorial nits. Please mention if you plan to fix them or not. Fine w/ merge w/o them being fixed, but reserve the right to clean up the text later on (from an editorial perspective).

- Changed the opening sentence to be neutral (inside of negative).
- Removed one occurrence of "DIDs" in the concluding sentence.  Prefer to leave it as a sentence (vs. a bulleted list).
- In `terms.html`, added the following sentence with link back to this section...
> A DID is a <a href="https://w3c-ccg.github.io/did-spec/#uris-urls-and-urns">URL, URI, and a URN</a>.

terms.html
```diff
@@ -27,7 +27,8 @@
 registered with <a>distributed ledger</a> technology or other form of
 decentralized network. The generic format of a DID is defined in this
 specification. A specific <a>DID scheme</a> is defined in a
-<a>DID method</a> specification.
+<a>DID method</a> specification. 
+A DID is a <a href="https://w3c-ccg.github.io/did-spec/#uris-urls-and-urns">URL, URI, and a URN</a>.
```
msporny: DIDs are *not* URNs. URNs start with the text "urn:" -- https://tools.ietf.org/html/rfc8141#section-2

index.html
```diff
@@ -286,6 +286,10 @@ <h2>
 changes controllers, etc. URNs are intended to last forever.
           </li>
         </ol>
+        <p>
+DIDs are valid URIs; more specifically, they are also valid URLs and URNs.
```
msporny: DIDs are *not* URNs. URNs start with the text "urn:" -- https://tools.ietf.org/html/rfc8141#section-2

mwherman2000:
@dmitrizagidulin There's technical details/differences that matter.  For example, https://github.com/w3c-ccg/did-spec/issues/97

dmitrizagidulin:
@mwherman2000 
> There's technical details/differences that matter.

I'm not seeing it.. Issue #97 has nothing to do with whether DIDs are URLs or URNs, it has to do with JSON-LD parsing and serialization mechanics.

But anyways, hopefully the issue has been settled? That DIDs are emphatically _not_ URNs, they do not start with `urn:`, and do not match the URN RFC?

terms.html
mwherman2000: @msporny Technically, you're correct: DIDs are not URNs according to RFC8141.

...and to top things off, DIDs might not even be URIs according to https://tools.ietf.org/html/rfc3986#section-1.2.2 ...quoting...

> 1.2.2.  Separating Identification from Interaction
> 
>    A common misunderstanding of URIs is that they are only used to refer
>    to accessible resources.  The URI itself only provides
>    identification; access to the resource is neither guaranteed nor
>    implied by the presence of a URI.  Instead, any operation associated
>    with a URI reference is defined by the protocol element, data format
>    attribute, or natural language text in which it appears.
> 
>    Given a URI, a system may attempt to perform a variety of operations
>    on the resource, as might be characterized by words such as "access",
>    "update", "replace", or "find attributes".  Such operations are
>    defined by the protocols that make use of URIs, not by this
>    specification.  However, we do use a few general terms for describing
>    common operations on URIs.  URI "resolution" is the process of
>    determining an access mechanism and the appropriate parameters
>    necessary to dereference a URI; this resolution may require several
>    iterations.  To use that access mechanism to perform an action on the
>    URI's resource is to "dereference" the URI.

This ties in with the comments in: https://github.com/w3c-ccg/did-spec/issues/121

I think we need decide whether a "DID" is just the identifier part (with a separate access protocol spec) so that we can call them URIs and treat them like URIs ...else we in a green field scenario and have to define everything and not reference the URI/URL/URN specs?

mwherman2000:
> @mwherman2000
> 
> > There's technical details/differences that matter.
> 
> I'm not seeing it.. Issue #97 has nothing to do with whether DIDs are URLs or URNs, it has to do with JSON-LD parsing and serialization mechanics.
> 
> But anyways, hopefully the issue has been settled? That DIDs are emphatically _not_ URNs, they do not start with `urn:`, and do not match the URN RFC?

DIDs might not even be URIs ... see https://github.com/w3c-ccg/did-spec/pull/159#discussion_r251186233

terms.html
msporny: > I think we need decide whether a "DID" (in the context of the draft DID spec) is just the identifier part (with a separate access protocol spec) so that we can call DIDs "URI"s and treat them like URIs

That was decided a while ago (although the spec text most likely does not reflect it accurately yet)...

1. Decentralized Identifiers are just the identifier. 
2. DIDs refer to DID Documents, which are serializations of the Decentralized Identifier Document Data Model, full stop.
3. There are protocols for interacting w/ DIDs (defined in DID Method specifications).

The DID Spec at present, defines 1, 2, and talks about what 3 has to elaborate upon. These things may be split out into separate specifications in time, but at present, they're in the same specification because we didn't want people to have to jump around too much in the early days to see how all of the pieces fit together.

msporny: >  the way the current draft DID spec is written/scoped, might not even be URIs

We should make it clear that there is a URI Definition part of the spec, a data model part of the spec, a serializations part of the spec, and an "everything else" part of the spec.

mwherman2000: This is headed in the right direction ...how do we get to some point of closure?

msporny: > This is headed in the right direction ...how do we get to some point of closure?

PRs is how we get to closure. Suggest a set of PRs to reorganize the spec into these sections. I expect @rhiaro could probably put something like that together... or I can do it when I get a next pass at the spec (which is going to be end of Feb based on all of the other fires that are being dealt with at present). These are editorial changes, so I expect they can go in w/o much debate.

Once there is a PR in place, we debate it in the PR, and then pull it in the spec. Sometimes the Editors just make the change if they think it's purely editorial and won't cause community objections. This feels like it falls into that category. So, a couple of ways forward... in this order of preference:

1. @rhiaro writes the PR.
2. @mwherman2000 writes the PR.
3. I write the PR.

mwherman2000: I'm going to have to bow out of this PR.  I've been banging on the DID spec and the INDY ARM for over 2 months now (self-funded). I'm trying to focus on the next layers of the Indy Platform e.g. the SDK (https://hyperonomy.com/2019/01/24/indy-sdk-getting_started-py-visualizations-light-table/).
I can still provide feedback ...very passionate about making this spec valuable for developers.

msporny: > I'm going to have to bow out of this PR.

No problem, you've given us enough of a direction to head in where we can convert this PR into something that'll make it into the spec. Thank you @mwherman2000 ... @rhiaro and I will discuss how we can get your input integrated into the spec over the next couple of weeks.

BigBlueHat:
I'd avoid the [WHATWG URL spec](https://url.spec.whatwg.org/#urls). It's intended for browser/JavaScript parsing of values in `src` and `href`'s. It's not a general foundation for identification like [RFC 3986](https://tools.ietf.org/html/rfc3986) (and related RFCs). I'd avoid it, unless you're building an HTML parser or JS runtime.

rhiaro:
Spec restructuring and terminology clarification is progressing, and I think the various discussions in this thread are resolved enough to close this PR now.

PR #162, opened 2019-01-26
Clarify PubKey section

rhiaro:
Cherry picks from https://github.com/w3c-ccg/did-spec/pull/61 to incorporate feedback in the discussion. 

Ready to merge now, but I have an additional proposal to clarify the "vague" text around value properties (https://github.com/w3c-ccg/did-spec/pull/61#discussion_r178646041), which is to move anything around key formats to 5., ie:

> 3. Each public key MUST include `id` and `type` properties.
> 4. ...
> 5. Each public key MUST contain exactly one key with a value of one of: `publicKeyPem`, `publicKeyJwk`, `publicKeyHex`, `publicKeyBase64`, `publicKeyBase58`, or similar depending on the format and encoding of the public key. The array of public keys SHOULD NOT contain duplicate entries with the same `id` and different formats.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/162.html" title="Last updated on Jan 26, 2019, 8:38 PM UTC (ed66265)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/162/dfdac17...ed66265.html" title="Last updated on Jan 26, 2019, 8:38 PM UTC (ed66265)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1122,39 +1122,39 @@ detail how revocation is performed and tracked.
 The rules for public keys are:
       </p>
 
-      <ol start="1">
-        <li>
-A DID Document MAY include a <code>publicKey</code> property.
-        </li>
+<ol start="1">
+  <li>
+A DID Document MAY include a <code>publicKey</code>
+property.
+  </li>
 
-        <li>
-The value of the <code>publicKey</code> property should be an
-array of public keys.
-        </li>
+  <li>
+The value of the <code>publicKey</code> property MUST be an
+array of public keys. 
+  </li>
 
-        <li>
-Each public key must include <code>id</code> and
-        <code>type</code> properties, and exactly one value property.
-        </li>
+  <li>
+Each public key MUST include <code>id</code> and
+<code>type</code> properties, and exactly one value property. The array of public keys SHOULD NOT contain duplicate entries
+    with the same <code>id</code> and different value properties with different formats.
+  </li>
 
-        <li>
-Each public key may include an <code>controller</code> property, which
-identifies the entity that controls the corresponding private key. If
-this property is missing, it is assumed to be the DID subject.
-        </li>
+  <li>
+Each public key MUST include a <code>controller</code> property, which identifies the entity that controls
+the corresponding private key. If this property is missing, it is assumed to be the DID subject.
+  </li>
 
-        <li>
-The value property of a public key may be
-        <code>publicKeyPem</code>, <code>publicKeyJwk</code>,
-        <code>publicKeyHex</code>, <code>publicKeyBase64</code> or similar,
-depending on the format and encoding of the public key.
-        </li>
+  <li>
+The value property of a public key MUST be exactly one of
+<code>publicKeyPem</code>, <code>publicKeyJwk</code>, <code>publicKeyHex</code>, <code>publicKeyBase64</code>, <code>publicKeyBase58</code>, or similar
+depending on the format and encoding of the public key. 
+  </li>
 
-        <li>
-A registry of key types and formats is available in Appendix
-<a href="#registries"></a>.
-        </li>
-      </ol>
+  <li>
+A registry of key types and formats is available in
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1128,30 +1128,33 @@ <h2>
...
+identifies the entity that controls the corresponding private key. If 
 this property is missing, it is assumed to be the DID subject.
         </li>
 
         <li>
-The value property of a public key may be
-        <code>publicKeyPem</code>, <code>publicKeyJwk</code>,
-        <code>publicKeyHex</code>, <code>publicKeyBase64</code> or similar,
-depending on the format and encoding of the public key.
+The value property of a public key MUST be exactly one of
```
rhiaro: Actually this should be `MAY`?

AxelNennker: No, MUST is correct, otherwise the spec would be open for implementers to use whatever and the result would be a break of interoperability

rhiaro: It's just kind of a weird MUST because the list is still open ended (I kept "or similar depending on the format and encoding of the public key" because the feedback in the other PR gave me the impression that was wanted).

index.html
```diff
@@ -1128,30 +1128,33 @@ <h2>
...
+<code>publicKeyPem</code>, <code>publicKeyJwk</code>, 
+<code>publicKeyHex</code>, <code>publicKeyBase64</code>, 
+<code>publicKeyBase58</code>, or similar depending on the format and
```
msporny: We will want to add "publicKeyMultibase" ... which might enable us to get rid of publicKeyHex, publicKeyBase64, and publicKeyBase58.

index.html
AxelNennker: I don't see a problem with upgradability of the spec. If the community decides that a new format warrants inclusion in the list of public key formats then there can either be a new version of this spec or another spec stating that a new public key format is needed. The same is true the other way round. If the community decides that a cryptographic method is not safe enough any more then that public key format might be recommended NOT to be used or removed from the list.  

A definitive list helps developers because it tells us what is reasonable to implement.
And it protects developers from big players who might choose to add complicated to implement algorithms which only big groups can sustain. A too loose spec does only help big groups/companies.

AxelNennker:
Should the specification of the public key formats go into https://w3c-ccg.github.io/ld-cryptosuite-registry/ ?
They are not there and there is not issue to add them.

Still, I suggest to remove the 'or similar' from the public key section. Reasons above

msporny:
> Should the specification of the public key formats go into https://w3c-ccg.github.io/ld-cryptosuite-registry/ ?

Yes, they should, good suggestion @AxelNennker.

> Still, I suggest to remove the 'or similar' from the public key section. Reasons above

Agreed. That makes the language tighter.

index.html
```diff
@@ -1128,30 +1128,33 @@ <h2>
...
+Each public key MUST include <code>id</code> and <code>type</code> 
+properties, and exactly one value property. The array of public keys 
+SHOULD NOT contain duplicate entries with the same <code>id</code> and 
+different value properties with different formats.
         </li>
 
         <li>
-Each public key may include an <code>controller</code> property, which
-identifies the entity that controls the corresponding private key. If
+Each public key MUST include a <code>controller</code> property, which
```
peacekeeper: If this is now a MUST, then why does the next sentence define what to do if the property is missing?

index.html
peacekeeper: Quick note on this older thread that per discussion in [2019-05-30 DID Spec and DID Resolution Spec meeting](https://github.com/w3c-ccg/meetings/tree/gh-pages/2019-05-30-did-spec-and-resolution), we are working on a new PR that will say that a public key property MUST be in the [Linked Data Cryptographic Suite Registry](https://w3c-ccg.github.io/ld-cryptosuite-registry/).

PR #165, opened 2019-02-05
Rename "cryptographic suites" to verification methods.

dlongley:
- A public key is a type of verification method, so remove
  nesting of public keys within what were previously termed
  "cryptographic suites" to simplify.


index.html
```diff
@@ -559,17 +560,13 @@ <h2>
 {
   "@context": "https://w3id.org/did/v1",
   "id": "did:example:123456789abcdefghi",
-  "publicKey": [{
+  "authentication": [{
+    // this key can be used to authenticate as DID ...fghi
```
davidlehn: This looks kind of like a typo (here and below).  Maybe `as DID "did:...fghi"`?

index.html
```diff
@@ -2590,26 +2586,25 @@ <h1>
...
     "publicKeyPem": "-----BEGIN PUBLIC KEY...END PUBLIC KEY-----\r\n"
-  }, {
-    "id": "did:example:123456789abcdefghi#keys-2",
-    "type": "Ed25519VerificationKey2018",
-    "controller": "did:example:123456789abcdefghi",
-    "publicKeyBase58": "H3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV"
   }, {
     "id": "did:example:123456789abcdefghi#keys-3",
-    "type": "RsaPublicKeyExchangeKey2018",
+    "type": "ieee2410VerificationKey2018",
```
davidlehn: Capitalize first letter of type:
```suggestion
    "type": "Ieee2410VerificationKey2018",
```

index.html
dlongley: Well, not my convention ... so I'm not going to change it in this PR. Could be changed later.

index.html
dlongley: Again, not my convention there -- just trying to keep things as close to how they were as before in this PR.

talltree:
I'm good with this.

PR #168, opened 2019-02-12
Refactor ABNF rules for DIDs and DID References.

dmitrizagidulin:
This PR integrates and reflects the current consensus on the DID ABNF rules, as well as DID Reference and Service Endpoint ABNF rules.

Specifically:

* Separates the DID Reference ABNF from the DID ABNF
* Differentiates a DID fragment from a Service fragment
* Specifies ABNF rules for did query, did path
* Clarifies use cases for DID References (service endpoints resolution, key references, etc). 
* Updates service endpoint `id:` properties in examples.

Replaces PR #106, addresses #85.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/168.html" title="Last updated on Feb 16, 2019, 6:54 PM UTC (89a234c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/168/aeb7bc6...89a234c.html" title="Last updated on Feb 16, 2019, 6:54 PM UTC (89a234c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -534,7 +534,7 @@ Document.
     "publicKeyPem": "-----BEGIN PUBLIC KEY...END PUBLIC KEY-----\r\n"
   }],
   "service": [{
-    "id": "#service123",
+    "id": "did:example:123456789abcdefghi#service123",
     "type": "ExampleService",
     "serviceEndpoint": "https://example.com/endpoint/8377464"
   }]
@@ -1423,31 +1423,31 @@ Example:
       <pre class="example nohighlight" title="Various service endpoints">
 {
   "service": [{
-    "id": "#openid",
+    "id": "did:example:123456789abcdefghi#openid",
     "type": "OpenIdConnectVersion1.0Service",
     "serviceEndpoint": "https://openid.example.com/"
   }, {
-    "id": "#vcr",
+    "id": "did:example:123456789abcdefghi#vcr",
     "type": "CredentialRepositoryService",
     "serviceEndpoint": "https://repository.example.com/service/8377464"
   }, {
-    "id": "#xdi",
+    "id": "did:example:123456789abcdefghi#xdi",
     "type": "XdiService",
     "serviceEndpoint": "https://xdi.example.com/8377464"
   }, {
-    "id": "#agent",
+    "id": "did:example:123456789abcdefghi#agent",
     "type": "AgentService",
     "serviceEndpoint": "https://agent.example.com/8377464"
   }, {
-    "id": "#hub",
+    "id": "did:example:123456789abcdefghi#hub",
     "type": "HubService",
     "serviceEndpoint": "https://hub.example.com/.identity/did:example:0123456789abcdef/"
   }, {
-    "id": "#messages",
+    "id": "did:example:123456789abcdefghi#messages",
     "type": "MessagingService",
     "serviceEndpoint": "https://example.com/messages/8377464"
   }, {
-    "id": "#inbox",
+    "id": "did:example:123456789abcdefghi#inbox",
     "type": "SocialWebInboxService",
     "serviceEndpoint": "https://social.example.com/83hfh37dj",
     "description": "My public social inbox",
@@ -1456,7 +1456,7 @@ Example:
       "currency": "USD"
     }
   }, {
-    "id": "#authpush",
+    "id": "did:example:123456789abcdefghi#authpush",
     "type": "DidAuthPushModeVersion1",
     "serviceEndpoint": "http://auth.example.com/did:example:123456789abcdefg"
   }]
@@ -1720,7 +1720,7 @@ context above and adding the new property to the <a>DID Document</a>.
   "authentication": [ ... ],
   "service": [<span class="highlight">{
     "@context": "did:example:contexts:987654321",
-    "id": "#photos",
+    "id": "did:example:123456789abcdefghi#photos",
     "type": "PhotoStreamService",
     "serviceEndpoint": "https://example.org/photos/379283"
   }</span>]
--- a/index.html
+++ b/index.html
@@ -819,7 +819,7 @@ The method name MUST be lowercase.
 
         <li>
 Case sensitivity and normalization of the value of the
-specific-idstring rule in Section <a href="#the-generic-did-scheme">
+<pre>method-specific-idstring</pre> rule in Section <a href="#the-generic-did-scheme">
           </a> MUST be defined by the governing DID method specification.
         </li>
       </ol>
@@ -1789,7 +1789,7 @@ DID Method Schemes
 
       <p>
 A DID method specification MUST define exactly one specific DID
-scheme identified by exactly one method name (the method rule in
+scheme identified by exactly one method name (the <pre>method</pre> rule in
 Section <a href="#the-generic-did-scheme"></a>). Since DIDs are
 intended for decentralized identity infrastructure, it is NOT
 RECOMMENDED to establish a registry of unique DID method names.
@@ -1813,18 +1813,19 @@ network to which the DID method specification applies.
 
       <p>
 The DID method specification for the specific DID scheme MUST specify
-how to generate the specific-idstring component of a DID. The
-specific-idstring value MUST be able to be generated without the use
-of a centralized registry service. The specific-idstring value SHOULD
-be globally unique by itself. The fully qualified DID as defined by
-the DID rule in Section <a href="#the-generic-did-scheme"></a> MUST
-be globally unique.
+how to generate the <pre>method-specific-idstring</pre> component of a DID. The
+<pre>method-specific-idstring</pre> value MUST be able to be generated without
+the use of a centralized registry service. The
+<pre>method-specific-idstring</pre> value SHOULD be globally unique by itself.
+The DID as defined by the <pre>did</pre> rule in Section
+<a href="#the-generic-did-scheme"></a> MUST be globally unique.
       </p>
 
       <p>
-If needed, a specific DID scheme MAY define multiple specific
-specific-idstring formats. It is RECOMMENDED that a specific DID
-scheme define as few specific-idstring formats as possible.
+If needed, a specific DID scheme MAY define multiple
+<pre>method-specific-idstring</pre> formats. It is RECOMMENDED that a specific
+DID scheme define as few <pre>method-specific-idstring</pre> formats as
+possible.
       </p>
     </section>
 
--- a/index.html
+++ b/index.html
@@ -778,7 +778,7 @@ service-id         = ALPHA / DIGIT / "." / "-" / "_"
 
       <section>
         <h3>
-      DID Reference Service Paths, Queries and Fragments
+      DID Reference Service Paths, Queries, and Fragments
         </h3>
 
       <p>
--- a/index.html
+++ b/index.html
@@ -736,7 +736,9 @@ portable way to refer to Service Endpoint resources, queries and fragments.
       </p>
 
       <pre class="nohighlight">
-did-reference      = did [ "#" did-fragment ] [";" service-id] [ "/" service-path ] [ "?" service-query ] [ "#" service-fragment ]
+did-reference      = did (did-fragment-reference / did-service-reference)
+did-fragment-reference  = "#" did-fragment
+did-service-reference   = ";" service-id] [ "/" service-path ] [ "?" service-query ] [ "#" service-fragment ]
 service-id         = ALPHA / DIGIT / "." / "-" / "_"
       </pre>
 
--- a/index.html
+++ b/index.html
@@ -730,53 +730,63 @@ idchar             = ALPHA / DIGIT / "." / "-"
 DID Reference
       </h2>
 
+      <p>DID References are used for two main purposes: to refer to sections
+inside a DID Document (via <a href="#did-fragments">DID Fragments</a>), and as a
+portable way to refer to Service Endpoint resources, queries and fragments.
+      </p>
+
       <pre class="nohighlight">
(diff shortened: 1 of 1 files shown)


dmitrizagidulin:
ABNF rules refactored. What remains (this should probably go in a separate PR? I'm not sure):

* Add an explanation for what DID Reference Service Paths are _for_
* Link to the corresponding section in the DID Resolution spec.

index.html
```diff
@@ -715,67 +716,85 @@ <h2>
...
-<a href="#the-generic-did-scheme"></a>) is identical to a URI
-fragment and MUST conform to the ABNF of the fragment ABNF rule in
-[[RFC3986]]. A DID fragment MUST be used only as a method-independent
-pointer into the DID Document to identify a unique key description or
-other DID Document component. To resolve this pointer, the complete
-DID reference including the DID fragment MUST be used as the value of
-the id key for the target JSON object.
-      </p>
+        <p>
+  A generic <a>DID fragment</a> (the <pre>did-fragment</pre> rule in Section
```
davidlehn: I think all the uses of `<pre>` in sentences in this patch should be using `<code>`?  The `<pre>` rendering is wrong.

index.html
```diff
@@ -715,67 +716,85 @@ <h2>
       </p>
```
davidlehn: As long as ABNF is being worked on, it seems weird to have the above loosely defining `ALPHA`.  If we're going to define `ALPHA` why not just put it below?  `DIGIT` should get the same treatment either way.  I'm not sure what the best practice is for this. Could maybe say:
```
The following is the ABNF definition using the syntax in [[RFC5234]]
which defines <code>ALPHA</code> and <code>DIGIT</code>.
```

index.html
dmitrizagidulin: Whoops.. Fixing..

index.html
```diff
@@ -715,67 +716,85 @@ <h2>
...
-Paths
+DID Reference
       </h2>
 
-      <p>
-A generic <a>DID path</a> (the did-path rule in Section <a href="#the-generic-did-scheme"></a>) is identical to a URI path and MUST
-conform to the ABNF of the path-rootless ABNF rule in [[RFC3986]]. A
-DID path SHOULD be used to address resources available via a DID
-service endpoint. See Section <a href="#service-endpoints"></a>.
+      <p>DID References are used for two main purposes: to refer to sections
```
davidlehn: style: start text on next line flush left.

dmitrizagidulin: Thx, fixed.

index.html
```diff
@@ -800,7 +820,7 @@ <h2>
 
         <li>
 Case sensitivity and normalization of the value of the
-specific-idstring rule in Section <a href="#the-generic-did-scheme">
+<code>method-specific-idstring</code> rule in Section <a href="#the-generic-did-scheme">
```
davidlehn: Remove the newline inside `<a ...></a>`.

dmitrizagidulin: Thx, fixed.

index.html
dmitrizagidulin: Fixed, thanks.

index.html
dmitrizagidulin: Thanks! Changed to your suggestion.

index.html
```diff
@@ -534,6 +534,7 @@ <h2>
     "publicKeyPem": "-----BEGIN PUBLIC KEY...END PUBLIC KEY-----\r\n"
   }],
   "service": [{
+    "id": "#service123",
```
davidlehn: use full URI

index.html
```diff
@@ -1402,31 +1423,31 @@ <h2>
       <pre class="example nohighlight" title="Various service endpoints">
 {
   "service": [{
-    "id": "did:example:123456789abcdefghi;openid",
+    "id": "#openid",
```
davidlehn: use full URI (here and below)

index.html
dmitrizagidulin: Fixed, thanks!

index.html
dmitrizagidulin: Fixed, thanks!

mwherman2000:
How soon can this PR be accepted?

index.html
```diff
@@ -702,80 +703,100 @@ <h2>
...
-      <p>
-A specific DID scheme MAY specify ABNF rules for DID paths that are
-more restrictive than the generic rules in this section.
-      </p>
-    </section>
+      <pre class="nohighlight">
+did-reference      = did (did-fragment-reference / did-service-reference)
+did-fragment-reference  = "#" did-fragment
+did-service-reference   = ";" service-id [ "/" service-path ] [ "?" service-query ] [ "#" service-fragment ]
+service-id         = ALPHA / DIGIT / "." / "-" / "_"
```
davidlehn: The following are not defined: `did-fragment`, `service-path`, `service-query`, `service-fragment`.

index.html
```diff
@@ -702,80 +703,100 @@ <h2>
...
+
+Additionally, a DID can be used as a component in a DID Reference, described in
+the <a href="#did-reference">DID Reference section</a> below.
       </p>
 
       <p>
-Following is the ABNF definition using the syntax in [[RFC5234]]
-        (which defines ALPHA as upper or lowercase A-Z).
+The following is the ABNF definition using the syntax in [[RFC5234]]
+which defines <code>ALPHA</code> and <code>DIGIT</code>.
```
davidlehn: Should this be moved up a section and phrased so that it covers The Generic DID Scheme section and DID Reference section?

davidlehn:
@dmitrizagidulin If the preview link is correct, there's a `<section>` tag issue somewhere.  Everything is under section 4, vs the current version where that is not the case.

davidlehn:
@dmitrizagidulin Does `terms.html` needs some updating too to match these changes?  For instance, the metions and links about "DID Path" .  (And while fixing that, maybe reword so the hard coded easy breakable "section x.y" text in that file isn't used)

index.html
```diff
@@ -702,80 +703,100 @@ <h2>
...
       </h2>
 
       <p>
-A generic <a>DID path</a> (the did-path rule in Section <a href="#the-generic-did-scheme"></a>) is identical to a URI path and MUST
-conform to the ABNF of the path-rootless ABNF rule in [[RFC3986]]. A
-DID path SHOULD be used to address resources available via a DID
-service endpoint. See Section <a href="#service-endpoints"></a>.
+DID References are used for two main purposes: to refer to sections
+inside a DID Document (via <a href="#did-fragments">DID Fragments</a>), and as a
+portable way to refer to Service Endpoint resources, queries and fragments.
```
davidlehn: oxford comma and link:
```suggestion
portable way to refer to <a>Service Endpoint</a> resources, queries, and fragments.
```

index.html
```diff
@@ -702,80 +703,100 @@ <h2>
...
+  given by the target URL without needing to rely on graph-based processing.
+        </p>
+        <p>
+  Implementations SHOULD NOT prevent the use of JSON pointers ([[RFC6901]]).
+        </p>
+      </section>
+
+      <section>
+        <h3>
+      DID Reference Service Paths, Queries and Fragments
```
davidlehn:
```suggestion
      DID Reference Service Paths, Queries, and Fragments
```

davidlehn:
There are more places where the def `<a>term</a>` style could be used. Unsure how far that sort of linking needs to go.

index.html
dmitrizagidulin: So, that's intentional. `did-fragment`, `-path` and `-query` are mentioned in the next paragraph (basically, the ABNF for them is the same as in the URI RFC).

davidlehn: Ah, I see.  Unsure if that needs to be noted out or not.  Leave it until someone else gets confused too.

dmitrizagidulin: Heh, yeah, I was wondering that same thing, but wasn't sure how to make it clearer. Does ABNF syntax have comments? :)

index.html
```diff
@@ -702,80 +703,100 @@ <h2>
...
+  fragment and MUST conform to the ABNF of the fragment ABNF rule in
+  [[RFC3986]]. A DID fragment MUST be used only as a method-independent
+  pointer into the DID Document to identify a unique key description or
+  other DID Document component. To resolve this pointer, the complete
+  DID reference including the DID fragment MUST be used as the value of
+  the id key for the target JSON object.
+        </p>
+
+        <p>
+  A specific DID scheme MAY specify ABNF rules for DID fragments that
```
davidlehn: Should this be `A specific <a>DID method</a> MAY...`?

index.html
```diff
@@ -702,80 +703,100 @@ <h2>
...
+rule in [[RFC3986]]. The path is used to address resources available via a DID
+service endpoint. See Section <a href="#service-endpoints"></a>.
       </p>
 
       <p>
-It is desirable that we enable tree-based processing of DIDs that include
-DID fragments (which resolve directly within the DID document) to locate
-metadata contained directly in the DID document or the service resource
-given by the target URL without needing to rely on graph-based processing.
+A specific DID scheme MAY specify ABNF rules for DID Reference service paths
```
davidlehn: Should this be `A specific <a>DID method</a> MAY`?

index.html
```diff
@@ -702,80 +703,100 @@ <h2>
...
+
+      <section>
+        <h3>
+      DID Reference Service Paths, Queries, and Fragments
+        </h3>
 
       <p>
-A specific DID scheme MAY specify ABNF rules for DID fragments that
-are more restrictive than the generic rules in this section.
+A DID Reference Service path (the <code>service-path</code> rule above) is
```
davidlehn: To be consistent:
```suggestion
A DID Reference service path (the <code>service-path</code> rule in Section
<a href="#did-reference"></a>) is
```

index.html
```diff
@@ -702,80 +703,100 @@ <h2>
...
-DID fragments (which resolve directly within the DID document) to locate
-metadata contained directly in the DID document or the service resource
-given by the target URL without needing to rely on graph-based processing.
+A specific DID scheme MAY specify ABNF rules for DID Reference service paths
+that are more restrictive than the generic rules in this section.
       </p>
+
       <p>
-Implementations SHOULD NOT prevent the use of JSON pointers ([[RFC6901]]).
+A DID Reference Service <code>query</code> and <code>fragment</code> parts are
```
davidlehn: To be consistent:
```suggestion
The DID Reference service query and service fragment
(the <code>service-query</code> and <code>service-fragment</code> rules
in Section <a href="#did-reference"></a>) parts are
```

index.html
dmitrizagidulin: Probably, yeah

index.html
davidlehn: The ABNF RFC is linked a few lines up!  Comments: https://tools.ietf.org/html/rfc5234#section-3.9

dhh1128:
I do not understand why we feel the need to use a different delimiter to distinguish service fragments from other types of fragments. Why can't we just use / for all fragments of any type, and require that their ID properties be unique within the DID Document?

mwherman2000:
@dhh1128 I agree.  At the `did-url` grammar level, [we should focus on allowable syntax constructs - in support for the desired  capabilities we want `did-url`s to have](https://hyperonomy.com/2019/03/11/giving-grammars-written-with-abnf-notation-the-respect-they-deserve/) ...not try to encode all sorts of specific capabilities and functions into the syntax.

Examples:
1. IMO a bad use of syntax: https://github.com/WebOfTrustInfo/rwot8-barcelona/pull/164#issuecomment-470869225
2. IMO a better use of syntax: https://github.com/w3c-ccg/did-resolution/issues/32#issuecomment-471179669

The latter, at a syntax level, supports many levels of functional/capability innovation without having to change the syntax.

peacekeeper:
@dmitrizagidulin do you think everything in this PR has been addressed by https://github.com/w3c-ccg/did-spec/pull/189 (in fact that one even cherry-picked a few commits from here)? Or would you say this PR here still has content that needs to be merged?

kdenhartog:
I wanted to highlight the example of what I brought up on the DID call today about key-type matrix param being used as a method to specify a specific key-type within a grouping of key-types. For example, let's say in my DID-Doc I specify a grouping of `RSAVerificationKey2016` `RSAVerificationKey2017` and  `RSAVerificationKey2018` and call this grouping `RSAVerificationSuite` which is what the key-type would be listed in my DID Doc.

E.g. my DID Doc would be:

```
{
  "@context": ["https://w3id.org/did/v1", "https://w3id.org/security/v1"],
  ...
  "publicKey": [{
    "id": "did:example:123456789abcdefghi#keys-1",
    "type": "RsaVerificationSuite",
    "controller": "did:example:123456789abcdefghi",
  }],
  ...
}
```

Then when referencing to the public key for verification, I could use `did:example:123456789abcdefghi;key-type=RSAVerification2018#keys-1`

To suggest that the same key can be used with multiple key types.

If you were to use this for service-type it would be something like a grouping of services where a common set of services sit behind a single endpoint. So for example, I may run an Agent which supports DIDComm v1 and DIDCommv2 and I would specify the serialization of the object by using:

`did:example:123456789abcdefghi;key-type=DIDComm-JSON#agent`

or 

`did:example:123456789abcdefghi;key-type=DIDComm-CBOR#agent`

This is just an early example of where I think it may be used. I'm not sure if this indirection is useful, necessary, or neither. With that said, I want to take the time to agree with @dmitrizagidulin point raised that matrix params are not a method of optimizing resolver implementations. They should be used to identify a specific resource when the DID Doc contains metadata that describes a set.

dmitrizagidulin:
This PR is superceded by PR #189, so, it's ok to close this one.

Issue #119, opened 2018-12-18
[Overview] "Because DIDs reside on a distributed ledger" - not true?


mwherman2000:
In https://w3c-ccg.github.io/did-spec/#overview, it states....

> This design eliminates dependence on centralized registries for identifiers as well as centralized certificate authorities for key management—the standard pattern in hierarchical PKI (public key infrastructure). Because DIDs reside on a distributed ledger, each entity may serve as its own root authority—an architecture referred to as DPKI (decentralized PKI).

1. "Because DIDs reside on a distributed ledger" What is a DID?  If a DID is a character string identifier, then it doesn't live on the the ledger, the DID Document lives on the ledger.  See the following diagram...

[Hyperledger Indy/Sovrin Comprehensive Architecture Reference Model (INDY ARM) - latest version](https://hyperonomy.com/2019/01/13/hyperledger-indy-sovrin-comprehensive-architecture-reference-model-arm/) - bullets (12) thru (16) in both the diagram, Narration, and principles.

Reword/clarify (here and throughout the draft specification).

jandrieu:
Closing as we have moved this over to the DIDWG did-spec repo.

Issue #150, opened 2019-01-05
"distributed ledger" and "ledger" used randomly (in place of one-and-other)


mwherman2000:
Throughout the document, both "distributed ledger" and "ledger" are used randomly (in place of one-and-other).

Strong suggestion: On first use, "distributed ledger" should appear as "distributed ledger (ledger)".  All subsequent references should consistently refer to "ledger" ...only ...as a single word.

mwherman2000:
Side bar: Should (all forms) of "ledger" be changed to "journal"? ...serious request (that is not likely to be adopted).   Reference: https://www.linkedin.com/feed/update/urn:li:activity:6487812287863283712

In discussions about the future directions for the Indy Wallet, we've talked about the introduction of the `subledgers` (https://www.cliffsnotes.com/study-guides/accounting/accounting-principles-i/subsidiary-ledgers-and-special-journals/subsidiary-ledgers) ...which is potentially confuding with the way "ledger" is currently used in the Indy project.

In the real world, what we're calling a "ledger" in the Indy (and most blockchain) projects is actually a journal.  Referernce: https://keydifferences.com/difference-between-journal-and-ledger.html#ComparisonChart

jandrieu:
Interesting suggestion. I expect the use of ledger lacks the kinds of rigor we should aim for. However I'm not sure "journal" is better.

What we really mean is the decentralized data store of record. Or something like that. I hesitate to dive into bike shedding, but certainly "ledger" of any kind is a disservice to DIDs that don't use DLTs, and instead use other storage mechanisms like IPFS or the hypothetical "data" method (similar to the data scheme URLs).

Issue #149, opened 2019-01-05
[Authorization and Delegation] Is this sentence correct if there is no requirement to use a Ledger?


mwherman2000:
In https://w3c-ccg.github.io/did-spec/#authorization-and-delegation, it states...

> Since Authorization and Delegation are typically implemented by the ledger, each DID Method specification is expected to detail how authorization and delegation are performed for the ledger.

In https://w3c-ccg.github.io/did-spec/#terminology, it states...

> A globally unique identifier that does not require a centralized registration authority because it is registered with distributed ledger technology or other form of decentralized network

The implication of the second quoted sentence is that a distributed ledger (aka ledger) technology is not required to implement a system that conforms with this draft DID spec.  If so, why is ledger used in the first quoted sentence?

OR13:
Agree, an alternative:

Each DID Method _MUST_ define how authorization and delegation are implemented, including any necessary cryptographic operations.
 
The sentence could also simply removed.

PR #184, opened 2019-03-30
Add dfn of Decentralized Identifer Registry

rhiaro:
Use 'Decentralized Identifier Registry' (a type of Verifiable Data Registry) instead of "target system", "ledger" and "distributed ledger". Some references to "distributed ledger" remain but not in direct reference to DIDs, just in explanatory text.

(DIR was discussed at RWOT8.)

This should address #119, #149 and #150

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/184.html" title="Last updated on Mar 31, 2019, 2:49 PM UTC (980fdbc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/184/da1f220...980fdbc.html" title="Last updated on Mar 31, 2019, 2:49 PM UTC (980fdbc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -206,17 +206,18 @@ management</a>.
       </p>
 
       <p>
-Globally distributed ledgers, decentralized P2P networks, and other systems
-with similar capabilities, provide an opportunity for fully <a>decentralized
-identity management</a>. In a decentralized identity system, entities
-(in the sense of discrete identifiable units such as—but not limited to—people
-and organisations) are free to use any shared root of trust, with neither
-centralized authority nor a single point of failure.
-Combining decentralized identity systems enables any entity to create and
-manage their own identifiers on any number of distributed, independent roots of
-trust, an architecture referred to as
-<a href="https://github.com/WebOfTrustInfo/rebooting-the-web-of-trust/blob/master/final-documents/dpki.pdf">
-DPKI (decentralized PKI)</a>.
+The emergence of distributed ledger technology (DLT), sometimes
+referred to as blockchain technology, provides the opportunity for
+fully <a>decentralized identity management</a>. In a decentralized
+identity system, entities (in the sense of discrete identifiable units
+such as—but not limited to—people and organisations) are free to use any
+shared root of trust.
+Globally distributed ledgers, decentralized P2P networks, or other systems
+with similar capabilities, provide the means for managing a root of
+trust with neither centralized authority nor a single point of
+failure. In combination, DLTs and decentralized identity systems
+enable any entity to create and manage their own identifiers on any
+number of distributed, independent roots of trust.
       </p>
 
       <p>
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -206,18 +206,17 @@ <h2>
...
-identity system, entities (in the sense of discrete identifiable units
-such as—but not limited to—people and organisations) are free to use any
-shared root of trust.
-Globally distributed ledgers, decentralized P2P networks, or other systems
-with similar capabilities, provide the means for managing a root of
-trust with neither centralized authority nor a single point of
-failure. In combination, DLTs and decentralized identity systems
-enable any entity to create and manage their own identifiers on any
-number of distributed, independent roots of trust.
+Globally distributed ledgers, decentralized P2P networks, and other systems
```
msporny: I would rip out changes in lines 209-250 and put it in another PR -- it's going to make an otherwise non-controversial PR, controversial. That, or we'll have to wait for the discussion around the change to the abstract/intro to settle before pulling this in. If we don't have changes to 209-250, it's a fairly easy editorial change. (Note: I might have missed something on the last call that made this change acceptable by the group w/o discussion).

rhiaro: Mm okay I'll revert these and we can work on the overview section more separately

msporny:
I think this PR is good to go, @talltree and @peacekeeper can one of you two review this so we can get it into the spec.

terms.html
```diff
@@ -43,6 +43,14 @@
 System</a>, and most national ID systems.
   </dd>
 
+  <dt><dfn data-lt="DIR|decentralized identifier registry|decentralized identifier registries">Decentralized Identifier Registry</dfn> (DIR)</dt>
+
+  <dd>
+A role a system performs to mediate the creation, verification, updating, and
```
peacekeeper: Can we say "enable" instead of "mediate"? Just a detail, if not that's okay too.

rhiaro: "mediate" is the same language as the Verified Data Registry definition in the VC spec. Does anyone else have any strong feelings about this?

talltree:
I am good with this PR.

terms.html
peacekeeper: Okay nevermind, sounds good.

talltree:
I'm okay with this too.

peacekeeper:
Merging per decision in [2019-05-09 DID Spec and DID Resolution Spec meeting](https://github.com/w3c-ccg/meetings/tree/gh-pages/2019-05-09-did-spec-and-resolution).

Note: This is being merged primarily so we have a unified term ("Decentralized Identifier Registry") everywhere across the spec. The final term could still be changed via search&replace in a subsequent PR if desired; for example, the term "DID network" has also been proposed.

peacekeeper:
Fixes #119

Issue #127, opened 2018-12-20
[TERMINOLOGY] DID Method - overloading issues - Representations of DID Documents are written to the Ledger. ...not DIDs. ...misuse of "DID"


mwherman2000:
In https://w3c-ccg.github.io/did-spec/#terminology, it states.

>DID Method
>A definition of how a specific DID scheme can be implemented on a specific distributed ledger or network, including the precise method(s) by which DIDs and DID Documents can be read, written, and revoked.

1. RE: "including the precise method(s) by which DIDs and DID Documents can be read, written, and revoked."  
The use of  "DIDs", "DID Documents" and "DIDs and DID Documents" are being overloaded here and not being disambiguated ...leading to overloading and #confudsion.  A DID is an atrributed inside a DID Document.  Only DID Documents or representations of DID Documents are read, written, revoked on the Ledger.

2. Misuse of "DID" ...see https://github.com/w3c-ccg/did-spec/issues/121#issuecomment-455267182

Reference: [Hyperledger Indy/Sovrin Comprehensive Architecture Reference Model (INDY ARM) - latest version](https://hyperonomy.com/2019/01/13/hyperledger-indy-sovrin-comprehensive-architecture-reference-model-arm/) - bullets (12) thru (16) in both the diagram, Narration, and principles.

OR13:
It might help to add  ...including the precise method(s) by which DIDs are constructed, how DIDs are resolved to DID Documents, how DID Documents are created, updated, and revoked.

A related confusion is the concept of revoking a DID... to my knowledge this is always implemented by some update to the DID Document.  The spec seems to have the same ambiguity in this section:

https://w3c-ccg.github.io/did-spec/#delete-revoke

Maybe the spec should clarify that DIDs cannot be revoked, but a DID that resolves to a DID Document which is revoked, is considered revoked.

mwherman2000:
@OR13 ...misuse of "DID" ...see original comment.

mwherman2000:
https://github.com/w3c-ccg/did-spec/commit/dddc566ea6d3f9f4005b088bdc3dd312a5307734 doesn't address the specific wording issue highlighted at the beginning of this issue. The changes affect many other things but not the wording problem highlighted in this issue.  Please re-open this issue until it is resolved.

The key question is: what precisely is being "read, written, and revoked", etc.?  I don't believe it is the DID .

Suggestion: Change "DID and DID Document" to "DID Document"

rhiaro:
[...]

PR #185, opened 2019-03-30
Cleanup terms around "DID scheme"

rhiaro:
Use "DID URI scheme" and "DID method scheme" instead of "generic" and "specific" DID scheme, re: https://github.com/w3c-ccg/did-spec/issues/127#issuecomment-469200045

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/185.html" title="Last updated on Mar 30, 2019, 11:02 PM UTC (d69e2a6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/185/ce1130d...d69e2a6.html" title="Last updated on Mar 30, 2019, 11:02 PM UTC (d69e2a6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -333,12 +333,12 @@ Purpose of This Specification
       </h2>
 
       <p>
-The first purpose of this specification is to define the generic
-DID scheme and a generic set of operations on DID documents that can be
+The first purpose of this specification is to define the DID URI
+scheme and a generic set of operations on DID documents that can be
 implemented for any distributed ledger or network capable of
 supporting DIDs. The second purpose of this specification is to
 define the conformance requirements for a DID method
-specification—a separate specification that defines a specific DID
+specification—a separate specification that defines a DID method
 scheme and specific set of DID document operations for a specific
 distributed ledger or network.
       </p>
@@ -733,11 +733,11 @@ enables authentication of the DID subject.
 
     <section>
       <h2>
-The Generic DID Scheme
+The DID URI Scheme
       </h2>
 
       <p>
-The generic <a>DID scheme</a> is a URI scheme conformant with
+The <a>DID URI scheme</a> is a URI scheme conformant with
 [[RFC3986]]. It consists of a DID followed by an optional path and/or
 fragment. The term DID refers only to the identifier conforming to
 the did rule in the ABNF below; when used alone, it does not include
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -333,12 +333,12 @@ <h2>
...
       <p>
-The first purpose of this specification is to define the generic
-DID scheme and a generic set of operations on DID documents that can be
+The first purpose of this specification is to define the DID URI
+scheme and a generic set of operations on DID documents that can be
 implemented for any distributed ledger or network capable of
 supporting DIDs. The second purpose of this specification is to
 define the conformance requirements for a DID method
-specification—a separate specification that defines a specific DID
+specification—a separate specification that defines a DID method
```
msporny: Suggestions: Change all references to "DID method scheme" to just "DID method". Using "scheme" in both places may confuse folks. While it is true that a "DID method" is a scheme, most readers will miss the nuance and most likely use "DID URI scheme" and "DID method scheme" interchangeably.

rhiaro: That works in some places, but I'm not sure about the section 'DID Method Schemes' (L1919) and generally how we differentiate between when the spec is talking about the method-specific character string in the DID, and the DID method itself. Or maybe we don't need to?

msporny: Would replacing that with something like: "A DID method specification MUST define a method name and further restrictions to the did-method-idstring defined in the DID URI scheme"? 

The goal would be to effectively state: "There is one DID URI scheme, that's in this specification." The "DID URI scheme" may be further restricted by DID method specifications."

... or does that not address the text you're concerned about?

talltree: So while I think that generally works, I still think we need to define and use a term for what the previous spec called a "DID method scheme", i.e., the syntax (ABNF) defined by a DID method that restricts the generic DID scheme. 

How about we just call it a "DID method syntax"?

Also, I want to be able to continue to use the term "generic DID syntax" for this portion of this spec so that we can mirror the same language from [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt). Read these two paragraphs:
```
1.1.1.  Generic Syntax

   Each URI begins with a scheme name, as defined in Section 3.1, that
   refers to a specification for assigning identifiers within that
   scheme.  As such, the URI syntax is a federated and extensible naming
   system wherein each scheme's specification may further restrict the
   syntax and semantics of identifiers using that scheme.
   ...
   A parser of the generic URI syntax can parse any URI reference into
   its major components.  Once the scheme is determined, further
   scheme-specific parsing can be performed on the components.  In other
   words, the URI generic syntax is a superset of the syntax of all URI
   schemes.
```
So I would really like to use the same language to talk about the relationship of DIDs and DID methods, i.e., to be able to say:

> The DID generic syntax is a superset of the syntax of all DID method syntaxes.

I'm drafting the proposed new text for the Decentralized Identifiers section of the spec (in [this Google doc](https://docs.google.com/document/d/1qnDExIVjU5bYc601qUdLZIi9UAs1ojlHyKnVoz2zlLM/edit?usp=sharing)) to use this language. See what you think.

tplooker:
@talltree great doc, with the 'service-type' and 'key-type' generic did parameter names should the description state, selects a list of services/publicKeys from the did doc?

talltree:
@tplooker Yes, that was the conclusion on today's DID spec call. @peacekeeper has the action item to suggest new text for those two parameter descriptions. I'm sure he'd welcome suggestions from you on that.

rhiaro:
I think this PR is now superceded by https://github.com/w3c-ccg/did-spec/pull/187/

peacekeeper:
> I think this PR is now superseded by #187

I think that's partially true, but I feel this PR here still has some good thoughts and content around the "DID scheme" terminology that the other PR doesn't fully address yet.

msporny:
The DID TF moved to close this on the 2019-08-01 call, in preference of merging PR #189.

PR #187, opened 2019-04-19
Update ABNF to use matrix parameters

### LINKED ISSUE #85: Did Query

### LINKED ISSUE #90: Service endpoint identification matrix parameter

### LINKED ISSUE #170: Inconsistent ABNF and related definitions
TLDR: The ABNF and related logical elements of a DID are inconsistently defined.

The current pull request refactoring the ABNF rules does not address the issue I'm raising here.  https://github.com/w3c-ccg/did-spec/pull/168

My first point is that the charter isn't about defining a DID reference. It's about defining a DID. The current spec defines a DID as merely a component in a DID reference.

My second point is that a DID should include the entirety of the string that is presented, just like a URL includes the query part and and fragment part. https://tools.ietf.org/html/rfc3986 The current spec, with its definition of a DID as a component in a did-reference, a URI of the following form would NOT be a DID per the spec:
```
did:example:123456789abcdefghi#keys-1
```
Per the current spec (and the PR) this would be a DID-reference, not a DID. Only the first part is a DID. This is stated explicitly:

> The generic DID scheme is a URI scheme conformant with [RFC3986]. It consists of a DID followed
> by an optional path and/or fragment. **The term DID refers only to the identifier conforming to the 
> did rule in the ABNF below**; when used alone, it does not include a path or fragment. A DID that 
> may optionally include a path and/or fragment is called a DID reference.

Even if one would prefer to follow the notion that the DID is "only the identifier", then this statement might be correct if rewritten as
> The generic DID-reference scheme is a URI scheme conformant with [RFC3986]. It consists of a 
> DID followed by an optional path and/or fragment. 

However, this seems inconsistent with the last statement in that paragraph:
> A DID that may optionally include a path and/or fragment is called a DID reference.

This sentence is logically impossible. A "DID" that includes a path or fragment is NOT a DID. It is a did-reference. It *is* that, not that it is *called* that. Further, by the previous statements, DIDs may not optionally include a path and/or fragment.

This is also out of sync with current consensus in the way we speak of DIDs. As illustrated in the last paragraph, we consistently refer to did-references as DIDs. From Section 4.3:

>It is desirable that we enable tree-based processing of **DIDs that include DID fragments** (which 
> resolve directly within the DID document) to locate metadata contained directly in the DID 
> document or the service resource given by the target URL without needing to rely on 
> graph-based processing.

Again, if "DIDs" can contain DID fragments, then by the above (problematic) language and the ABNF, that "DID" is actually a DID reference which by definition can't be true given the current ABNF.

My third point is that the did-reference spec *looks* like the URI spec, but is fundamentally different.
From the current spec:
```abnf
did-reference      = did [ "/" did-path ] [ "#" did-fragment ]
did                = "did:" method ":" specific-idstring
```

However, this is significantly different from the generic URI syntax (RFC3986)
```abnf
URI         = scheme ":" hier-part [ "?" query ] [ "#" fragment ]
```
And further breaks with this URI-reference spec in RFC3986. It makes DID-references something other than URI-references, which makes DIDs something other than URIs.
```abnf
URI-reference = URI / relative-ref
```

A URI-reference is either a URI, or a relative-ref and a relative-ref has the following syntax:
```abnf
      relative-ref  = relative-part [ "?" query ] [ "#" fragment ]

      relative-part = "//" authority path-abempty
                    / path-absolute
                    / path-noscheme
                    / path-empty
```

If we attempt to follow that pattern more closely (yet combining the query & fragment parts to remove redundancy), I propose:
```abnf
did = "did:" method specific-idstring relative-part 
method             = 1*methodchar
methodchar         = %x61-7A / DIGIT
specific-idstring  = idstring *( ":" idstring )
idstring           = 1*idchar
idchar             = ALPHA / DIGIT / "." / "-"
relative-part = path-part [ "?" query] [ "#" fragment ]
path-part = / path-absolute
                    / path-empty
did-reference = DID / relative-part
```
[Yes, this is incomplete. I leave query and fragment as an exercise for the editors. Pull them in from the RFC3986]

With this definition, a DID-reference is a URI-reference where the query term is the name of a service used  for endpoint discovery by the user-agent. It is aligned with how URI-references are used. This implies that fragments should NOT be used to indicate dereferencing a DID to the resource at the end of a service endpoint, but rather to refer to the element itself (within the DID Document). Only the query term is interpreted as implying the further dereference.

Unfortunately, if service (from did-spec) is reasonably replaced with query (from the URI spec), then we still have the question of what the path part does. The current spec says 
> A DID path SHOULD be used to address resources available via a DID service endpoint.

If I follow the ideas of Sam Smith and DADs, then the path part would allow hierarchical indexing to provide resolution to an subset of a DID Document rather than the root of the DID Document. This would be a distinct from the query/service component which is for discovering the named service endpoint for a given DID id/path reference. Which is to say, I think this remains a point of significant ambiguity in the feature space (what is the path part)--and not just a grammatical error. I'm not sure I have a recommendation in this area. I don't know of any implementations that use such hierarchies other than what I recall from DADs--and I believe those may be addresses on a method-specific basis with colon ":" delimited identifiers.

Additionally, the section on fragments has this gem:
> It is desirable that we enable tree-based processing of DIDs that include DID fragments
> (which resolve directly within the DID document) to locate metadata contained directly 
> in the DID document or the service resource given by the target URL without needing to rely on graph-based processing.

This paragraph is ambiguous as to whether or not a fragment is a reference within the context of the document, as suggested by " (which resolve directly within the DID document)" or if it can reference something within the service endpoint, as suggested by "to locate metadata contained directly 
in the DID document **or the service resource**". I take this to mean that if the DID-reference (aka DID) contains a service (aka query), then any fragment reference is to be appended to the service endpoint for dereferencing in the context of that service endpoint. I can see the appeal of this. However, it presumes a certain understanding and stability of the structure of the named endpoint and could easily result in an invalid URI if the service endpoint already has a fragment. Further, it is doubly ambiguous when a fragment is used to refer to a service endpoint, as seems to be the proposal in the PR. 

Rather, I would use the query part to look up the service, and when dereferencing the service endpoint, and then applying the fragment in the context of the dereferenced endpoint at lease uses a valid URL (just one fragment).

For example, using Example 8 from the spec https://w3c-ccg.github.io/did-spec/#example-8-various-service-endpoints 
```
"service": [{
    "id": "did:example:123456789abcdefghi;openid",
    "type": "OpenIdConnectVersion1.0Service",
    "serviceEndpoint": "https://openid.example.com/"
  },
```
A relative part (query and fragment) of  **?did:example:123456789abcdefghi;openid** would dereference to the open id endpoint.

This is different from both the current spec--with its ambiguity around what the ";" is for (it is in the service section but not in the ABNF). And the PR mentioned above, which creates some new elements, which IMO are unnecessary and confusing as well as using fragments to refer to services.

Further, the current spec and proposed PR both use absolute URIs for service endpoint ids, but the PR says "the service path [is] a URI path and MUST conform to the ABNF of the path-rootless ABNF
rule in [[RFC3986]]" Unfortunately, this is logically impossible. It is either a URI path (which can be a path-abempty, path-absolute, path-noscheme, path-rootless, or path-empty), or it is a path-rootless.

I may have mis-parsed the ABNF, but the current use of the service endpoints is inconsistent with how fragments work in other URIs, especially in URLs. For the sake of this example, let's revisit the above example (from Section 8. Examples) and just focus on the id of the first service endpoint:

> "id": "did:example:123456789abcdefghi;openid"

which, in the PR is changed to 
> "id": "did:example:123456789abcdefghi#openid"

If we were to use a standard URL to refer to an element with that id in an HTML page (http://example.com), we would use either the absolute URL
`http://example.com#did:example:123456789abcdefghi;openid`
from the spec, or 
`http://example.com#did:example:123456789abcdefghi#openid`
from the PR.

If we apply that pattern to a service identifier in the DID Document resolved by `did:example:123456789abcdefghi`, then the analogous URL would be 
`did:example:123456789abcdefghi#did:example:123456789abcdefghi;openid`
for the original spec and
`did:example:123456789abcdefghi#did:example:123456789abcdefghi#openid`
for the PR.

Unfortunately, these PR URIs are invalid, because they contain two fragments, and all four are unnecessarily redundant.

In contrast, what I believe is the desired original DID is something like
```
did:example:123456789abcdefghi#openid
```
Which would refer to the openid service endpoint declaration in the DID Document and, based on my proposal, 
```
did:example:123456789abcdefghi?openid
```
Would actually dereference the service at the endpoint.

This is not only a valid URI, it's concise and straightforward.
In order to have that URI be consistent with how other URI fragments map to the ids in a document, 
the service endpoint should in the example 8 should be:
```
"service": [{
    "id": "openid",
    "type": "OpenIdConnectVersion1.0Service",
    "serviceEndpoint": "https://openid.example.com/"
  },
```

My comments therefore are in two parts:
1. The grammar of the current ABNF is inconsistent with itself, with URIs, and with how we talk about DIDs. This lack of rigor is **especially** problematic in ABNF which is there precisely to be rigorous because the rest of the document suffers from the inherent ambiguity in human language.
2. The logic of the current DID components are inconsistent with their similarly named parts in the URI spec. I propose that "path", "query", and "fragment" are better terms if they have slightly different definitions:
a. path: like a URI path, a hierarchical index into the DID document
b. query: the name of a service endpoint for discovery (and eventual dereferencing)
c. fragment: a named index into the DID document, hierarchically constrained by the path
3. Seems like a the query part should follow RFC3986 def of queries, not path-rootless.

My apologies this is so long.

I'd like @burnburn and @ChristopherA to comment on this.

### LINKED ISSUE #177: ABNF idchar should include underscore
To accommodate url-safe base64 encoded values, the idchar definition should be
ALPHA / DIGIT / "." / "-" / "_"

peacekeeper:
This updates the ABNF and introduces matrix parameters, based on discussions on the mailing list as well as the [DID Spec and DID Resolution Spec Weekly Meetings](https://docs.google.com/document/d/1qYBaXQMUoB86Alquu7WBtWOxsS8SMhp1fioYKEGCabE/).

Fixes #85. Fixes #90. Fixes #170. Fixes #177.
Replaces https://github.com/w3c-ccg/did-spec/pull/106.
Related to #185.

**[PR is still work-in-progress, do not merge yet!]**

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/187.html" title="Last updated on Apr 26, 2019, 4:57 PM UTC (9ec0477)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/187/ce1130d...9ec0477.html" title="Last updated on Apr 26, 2019, 4:57 PM UTC (9ec0477)">Diff</a>

### DIFF
--- a/common.js
+++ b/common.js
@@ -85,7 +85,7 @@ var ccg = {
       status: "CG-DRAFT",
       publisher: "Digital Verification Community Group"
     },
-    "DRAFT-SPORNY-HASHLINK-02": {
+    "HASHLINK": {
       title: "Cryptographic Hyperlinks",
       date: "December 2018",
       href: "https://tools.ietf.org/html/draft-sporny-hashlink-02",
--- a/common.js
+++ b/common.js
@@ -84,6 +84,25 @@ var ccg = {
       ],
       status: "CG-DRAFT",
       publisher: "Digital Verification Community Group"
+    },
+    "DRAFT-SPORNY-HASHLINK-02": {
+      title: "Cryptographic Hyperlinks",
+      date: "December 2018",
+      href: "https://tools.ietf.org/html/draft-sporny-hashlink-02",
+      authors: [
+        "Manu Sporny"
+      ],
+      status: "Internet-Draft",
+      publisher: "IETF"
+    },
+    "MATRIX-URIS": {
+      title: "Matrix URIs - Ideas about Web Architecture",
+      date: "December 1996",
+      href: "https://www.w3.org/DesignIssues/MatrixURIs.html",
+      authors: [
+        "Tim Berners-Lee"
+      ],
+      status: "Personal View"
     }
   }
 };
--- a/index.html
+++ b/index.html
@@ -898,7 +898,7 @@ may not be supported by all DID methods.
             </td>
             <td>
 A resource hash of the resource identified by the DID URL to add integrity
-protection, as specified in [[DRAFT-SPORNY-HASHLINK-02]].
+protection, as specified in [[HASHLINK]].
             </td>
           </tr>
 
--- a/index.html
+++ b/index.html
@@ -1062,7 +1062,7 @@ The method name MUST be lowercase.
 
         <li>
 Case sensitivity and normalization of the value of the
-specific-idstring rule in Section <a href="#generic-did-syntax">
+<code>method-specific-id</code> rule in Section <a href="#generic-did-syntax">
           </a> MUST be defined by the governing DID method specification.
         </li>
       </ol>
@@ -2171,9 +2171,9 @@ network to which the DID method specification applies.
 
       <p>
 The DID method specification for the specific DID scheme MUST specify
-how to generate the specific-idstring component of a DID. The
-specific-idstring value MUST be able to be generated without the use
-of a centralized registry service. The specific-idstring value SHOULD
+how to generate the <code>method-specific-id</code> component of a DID. The
+<code>method-specific-id</code> value MUST be able to be generated without the use
+of a centralized registry service. The <code>method-specific-id</code> value SHOULD
 be globally unique by itself. The fully qualified DID as defined by
 the DID rule in Section <a href="#generic-did-syntax"></a> MUST
 be globally unique.
@@ -2181,8 +2181,8 @@ be globally unique.
 
       <p>
 If needed, a specific DID scheme MAY define multiple specific
-specific-idstring formats. It is RECOMMENDED that a specific DID
-scheme define as few specific-idstring formats as possible.
+<code>method-specific-id</code> formats. It is RECOMMENDED that a specific DID
+scheme define as few <code>method-specific-id</code> formats as possible.
       </p>
     </section>
 
--- a/index.html
+++ b/index.html
@@ -971,12 +971,12 @@ illustrates both:
 
     <section>
       <h2>
-Paths
+Path
       </h2>
 
       <p>
-A generic <a>DID path</a> (the did-path rule in Section <a href="#the-generic-did-scheme"></a>) is identical to a URI path and MUST
-conform to the ABNF of the path-rootless ABNF rule in [[RFC3986]]. A
+A generic <a>DID path</a> is identical to a URI path and MUST
+conform to the the <code>path-abempty</code> ABNF rule in [[RFC3986]]. A
 DID path SHOULD be used to address resources available via a DID
 service endpoint. See Section <a href="#service-endpoints"></a>.
       </p>
@@ -989,13 +989,12 @@ more restrictive than the generic rules in this section.
 
     <section>
       <h2>
-Fragments
+Fragment
       </h2>
 
       <p>
-A generic <a>DID fragment</a> (the did-fragment rule in Section
-<a href="#the-generic-did-scheme"></a>) is identical to a URI
-fragment and MUST conform to the ABNF of the fragment ABNF rule in
+A generic <a>DID fragment</a> is identical to a URI
+fragment and MUST conform to the the <code>fragment</code> ABNF rule in
 [[RFC3986]]. A DID fragment MUST be used only as a method-independent
 reference into the DID Document to identify a component of a DID Document
 (e.g. a unique key description). To resolve this reference, the complete
--- a/index.html
+++ b/index.html
@@ -739,11 +739,17 @@ Generic DID Syntax
       <p>
 The generic <a>DID scheme</a> is a URI scheme conformant with
 [[RFC3986]]. The DID scheme specializes only the scheme and
-authority components of a DID URI—the path-abempty, query,
-and fragment components are identical to the ABNF rules defined
-in [[RFC3986]]. Note that the term “DID” refers only to the URI
-conforming to the did rule in the ABNF below. A DID always identifies
-the DID subject. The term “DID URL”, defined by the did-url rule,
+authority components of a DID URI—the <code>path-abempty</code>,
+<code>query</code>, and <code>fragment</code> components are
+identical to the ABNF rules defined
+in [[RFC3986]].
+      </p>
+
+      <p>
+The term <a>DID</a> refers only to the URI
+conforming to the <code>did</code> rule in the ABNF below. A
+DID always identifies the DID subject. The term <a>DID URL</a>,
+defined by the <code>did-url</code> rule,
 refers to a URL that begins with a DID followed by one or more
 additional components. A DID URL always identifies the resource to
 be located.
@@ -790,12 +796,11 @@ Generic DID Parameter Names
 
       <p>
 DID URL syntax supports a simple, generalized format for parameters based on the
-matrix parameter syntax ([[MatrixURIs]]) originally proposed in 1996 by Tim
-Berners-Lee and his collaborators. 
+matrix parameter syntax ([[MatrixURIs]]).
 The ABNF above does not specify any parameter names (the <code>param-name</code>
 rule). The following table defines a set of generic DID parameter names for
 parameters that MUST operate uniformly across all DID methods. (Method-specific
-parameter names are covered in the following section.)
+parameter names are covered in <a href="#method-specific-did-parameter-names"></a>.)
       </p>
 
       <table class="simple">
@@ -905,7 +910,7 @@ The exact processing rules for these parameters are specified in the DID
 Resolution specification.
       </p>
 
-      <p>
+      <p class="note">
 Note that there may be additional parameters or options that are not part
 of the DID URL but instead passed to a DID resolver “out of band”, i.e.,
 using a resolution protocol or some other mechanism. Such options could
@@ -940,7 +945,7 @@ this method and this method-specific parameter would be:
       </p>
 
       <p>
-Note that a method-specific parameter name defined by one DID method MAY
+A method-specific parameter name defined by one DID method MAY
 be used by other DID methods. For example:
       </p>
 
@@ -958,7 +963,7 @@ names in any order.
       </p>
 
       <p>
-Note that both DID method namespaces and method-specific parameter
+Both DID method namespaces and method-specific parameter
 namespaces may include colons, so they may be partitioned hierarchically
 as defined by a DID method specification. Here is an example DID URL that
 illustrates both:
--- a/index.html
+++ b/index.html
@@ -1021,7 +1021,7 @@ fragment and MUST conform to the the <code>fragment</code> ABNF rule in
 [[RFC3986]]. A DID fragment MUST be used only as a method-independent
 reference into the DID Document to identify a component of a DID Document
 (e.g. a unique key description). To resolve this reference, the complete
-DID reference including the DID fragment MUST be used as the value of
+DID URL including the DID fragment MUST be used as the value of
 the key for the target component in the DID Document object.
       </p>
 
--- a/index.html
+++ b/index.html
@@ -796,7 +796,7 @@ Generic DID Parameter Names
 
       <p>
 DID URL syntax supports a simple, generalized format for parameters based on the
-matrix parameter syntax ([[MatrixURIs]]).
+matrix parameter syntax ([[MATRIX-URIS]]).
 The ABNF above does not specify any parameter names (the <code>param-name</code>
 rule). The following table defines a set of generic DID parameter names for
 parameters that MUST operate uniformly across all DID methods. (Method-specific
@@ -898,7 +898,7 @@ may not be supported by all DID methods.
             </td>
             <td>
 A resource hash of the resource identified by the DID URL to add integrity
-protection, as specified in [[Hashlink]].
+protection, as specified in [[DRAFT-SPORNY-HASHLINK-02]].
             </td>
           </tr>
 
--- a/index.html
+++ b/index.html
@@ -745,7 +745,7 @@ identical to the ABNF rules defined
 in [[RFC3986]].
       </p>
 
-      <p>
+      <p class="note">
 The term <a>DID</a> refers only to the URI
 conforming to the <code>did</code> rule in the ABNF below. A
 DID always identifies the DID subject. The term <a>DID URL</a>,
--- a/index.html
+++ b/index.html
@@ -798,114 +798,114 @@ parameters that MUST operate uniformly across all DID methods. (Method-specific
 parameter names are covered in the following section.)
       </p>
 
-        <table class="simple">
-          <thead>
-            <tr>
-              <th>Parameter Name</th>
-              <th>Description</th>
-            </tr>
-          </thead>
+      <table class="simple">
+        <thead>
+          <tr>
+            <th>Parameter Name</th>
+            <th>Description</th>
+          </tr>
+        </thead>
 
-          <tbody>
-            <tr>
-              <td>
+        <tbody>
+          <tr>
+            <td>
 <code>service</code>
-              </td>
-              <td>
+            </td>
+            <td>
 Selects a <code>service</code> JSON-LD object from the DID Document by id.
-              </td>
-            </tr>
+            </td>
+          </tr>
 
-            <tr>
-              <td>
+          <tr>
+            <td>
 <code>service-type</code>
-              </td>
-              <td>
+            </td>
+            <td>
 Selects all <code>service</code> JSON-LD objects from the DID Document by type.
-              </td>
-            </tr>
+            </td>
+          </tr>
 
-            <tr>
-              <td>
+          <tr>
+            <td>
 <code>key</code>
-              </td>
-              <td>
+            </td>
+            <td>
 Selects a <code>publicKey</code> JSON-LD object from the DID Document by id.
-              </td>
-            </tr>
+            </td>
+          </tr>
 
-            <tr>
-              <td>
+          <tr>
+            <td>
 <code>key-type</code>
-              </td>
-              <td>
+            </td>
+            <td>
 Selects all <code>publicKey</code> JSON-LD objects from the DID Document by type.
-              </td>
-            </tr>
+            </td>
+          </tr>
 
-            <tr>
-              <td>
+          <tr>
+            <td>
 <code>version-id</code>
-              </td>
-              <td>
+            </td>
+            <td>
 Specifies a specific version of a DID Document to be resolved (the version ID
 could be sequential, or a UUID, or method-specific). Note: This parameter may
 not be supported by all DID methods.
-              </td>
-            </tr>
+            </td>
+          </tr>
 
-            <tr>
-              <td>
+          <tr>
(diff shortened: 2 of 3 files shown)


index.html
```diff
@@ -733,35 +733,239 @@ <h1>
...
+DID URL syntax supports a simple, generalized format for parameters based on the
+matrix parameter syntax ([[MatrixURIs]]) originally proposed in 1996 by Tim
+Berners-Lee and his collaborators. 
+The ABNF above does not specify any parameter names (the <code>param-name</code>
+rule). The following table defines a set of generic DID parameter names for
+parameters that MUST operate uniformly across all DID methods. (Method-specific
+parameter names are covered in the following section.)
+      </p>
+
+        <table class="simple">
```
davidlehn: Formatting issue: all the tags down to next `section` should shift left 2 spaces.

peacekeeper: Thanks I just fixed that.

peacekeeper:
Here's an experimental parser generated from the proposed new ABNF:
https://parser.uniresolver.io/

Source code:
https://github.com/decentralized-identity/did-common-java/tree/matrix-parameters

index.html
```diff
@@ -733,46 +733,250 @@ <h1>
...
+
+    <section>
+      <h2>
+Generic DID Parameter Names
+      </h2>
+
+      <p>
+DID URL syntax supports a simple, generalized format for parameters based on the
+matrix parameter syntax ([[MatrixURIs]]) originally proposed in 1996 by Tim
+Berners-Lee and his collaborators.
```
rhiaro: Editorial comment: "originally proposed in 1996 by Tim Berners-Lee and his collaborators" seems like background or primer type text that doesn't need to be in the spec. Sufficient to end this sentence after [[MatrixURIs]].

index.html
```diff
@@ -733,46 +733,250 @@ <h1>
...
+      </h2>
+
+      <p>
+DID URL syntax supports a simple, generalized format for parameters based on the
+matrix parameter syntax ([[MatrixURIs]]) originally proposed in 1996 by Tim
+Berners-Lee and his collaborators. 
+The ABNF above does not specify any parameter names (the <code>param-name</code>
+rule). The following table defines a set of generic DID parameter names for
+parameters that MUST operate uniformly across all DID methods. (Method-specific
+parameter names are covered in the following section.)
```
rhiaro: Editorial: s/`the following section`/`<a href="#method-specific-did-parameter-names"></a>`

index.html
```diff
@@ -733,46 +733,250 @@ <h1>
...
+Generic DID Parameter Names
+      </h2>
+
+      <p>
+DID URL syntax supports a simple, generalized format for parameters based on the
+matrix parameter syntax ([[MatrixURIs]]) originally proposed in 1996 by Tim
+Berners-Lee and his collaborators. 
+The ABNF above does not specify any parameter names (the <code>param-name</code>
+rule). The following table defines a set of generic DID parameter names for
+parameters that MUST operate uniformly across all DID methods. (Method-specific
```
yancyribbens: If DID methods can't or won't implement some of the proposed parameters because of a preference for greater decentralization by limiting the heavy lifting done by resolvers, then there should be the option to not implement.  Prefer wording that says _if_ a DID method implements a parameter, that DID method parameter MUST operate uniformly.

yancyribbens: If DID methods can't or won't implement some of the proposed parameters because of a preference for greater decentralization by limiting the heavy lifting done by resolvers, then there should be the option to not implement. Prefer wording that says if a DID method implements a parameter, that DID method parameter MUST operate uniformly.

index.html
```diff
@@ -733,46 +733,250 @@ <h1>
...
 The generic <a>DID scheme</a> is a URI scheme conformant with
-[[RFC3986]]. It consists of a DID followed by an optional path and/or
-fragment. The term DID refers only to the identifier conforming to
-the did rule in the ABNF below; when used alone, it does not include
-a path or fragment. A DID that may optionally include a path and/or
-fragment is called a DID reference.
+[[RFC3986]]. The DID scheme specializes only the scheme and
+authority components of a DID URI—the path-abempty, query,
+and fragment components are identical to the ABNF rules defined
+in [[RFC3986]]. Note that the term “DID” refers only to the URI
```
rhiaro: Editorial suggestion: put this "Note" in a note box (`<p class="note">Note that... to be located</p>`) to break it apart from the normative text.

index.html
```diff
@@ -733,46 +733,250 @@ <h1>
...
+parameter may not be supported by all DID methods.
+            </td>
+          </tr>
+
+          <tr>
+            <td>
+<code>content-type</code>
+            </td>
+            <td>
+Specifies a type of content object (other than a DID Document) to be returned
```
rhiaro: Maybe I missed something - what other kinds of content objects are in scope (and not implementation-specific) to be returned from the DID registry, other than a DID Document?

index.html
```diff
@@ -733,46 +733,250 @@ <h1>
...
+
       <p>
-See Sections <a href="#paths"></a> and <a href="#fragments"></a> for the ABNF rules defining DID paths and
-fragments.
+The exact processing rules for these parameters are specified in the DID
+Resolution specification.
+      </p>
+
+      <p>
+Note that there may be additional parameters or options that are not part
```
rhiaro: Propose putting this whole paragraph in a `<p class="note">` block

index.html
```diff
@@ -733,46 +733,250 @@ <h1>
...
+the parameter name must be <code>foo:bar</code>. An example DID URL using
+this method and this method-specific parameter would be:
+      </p>
+
+      <p>
+<code>did:foo:21tDAKCERh95uGgKbJNHYp;foo:bar=high</code>
+      </p>
+
+      <p>
+Note that a method-specific parameter name defined by one DID method MAY
```
rhiaro: Editorial: remove "Note that"

index.html
```diff
@@ -733,46 +733,250 @@ <h1>
...
+Method-specific parameter names MAY be combined with generic parameter
+names in any order.
+      </p>
+
+      <p>
+<code>did:example:21tDAKCERh95uGgKbJNHYp;service=agent;foo:bar=high</code>
+      </p>
+
+      <p>
+Note that both DID method namespaces and method-specific parameter
```
rhiaro: Editorial: remove "Note that"

index.html
peacekeeper: I accepted all your suggested edits except for this one. Instead I thought I'd remove the "Note that" and make it normative text. Do you agree, or would you rather see this in a note box?

index.html
peacekeeper: At RWoT#8 (I wasn't there), the idea of ["content references"](https://github.com/WebOfTrustInfo/rwot8-barcelona/blob/master/topics-and-advance-readings/DID-Content-References.md) came up, which means that a DID URL could refer to some object in the registry other than a DID Document. For example, in the Sovrin/Indy universe, there is a use case to refer to schemas and credential definitions. At RWoT#8 the idea was to propose additional syntax to accomplish this, but now we thought we can also just use matrix parameters to identify such additional objects. I'm not convinced that the concrete parameters that are currently here (`content-type`, `content-id`) are ideal. For example, perhaps path or query or fragment could be used instead of `content-id` or something like that. I would propose however that we don't let questions about individual matrix parameters stand in the way of merging the PR - we can always fine-tune the exact parameters later.

index.html
rhiaro: If this is to be normative text (I'm sure how it'd be testable though) I think it belongs in the terminology section of the spec rather than here. I think it works fine here as a reminder note, but seems not quite on-topic as a normative statement.

index.html
rhiaro: +1 this reads like all parameters are required by all DID methods. Should be an IF-MUST. Suggest to change

> The following table defines a set of generic DID parameter names for
> parameters that MUST operate uniformly across all DID methods.

to

> The following table defines a set of generic DID parameter names. DID
> methods MAY implement any or all of the parameters, and those 
> implemented MUST operate uniformly across all DID methods.

index.html
yancyribbens: > what other kinds of content objects are in scope..

+1.  Referencing other objects seems out of scope unless there is a specific use case.

> For example, in the Sovrin/Indy universe..

@peacekeeper This sounds like an implementation specific use case..  Any use cases that's specific to more than one implementation?

peacekeeper: I agree the schema / credential definition example is specific to only the "sov" DID method, but the concept itself (having a mechanism to reference arbitrary content in the target system) could be generic. @ken-ebert and @talltree brought up this requirement of "content references".

index.html
peacekeeper: Since all DID methods must be able to resolve a DID to a DID Document, and the first four parameters (`service`, `service-type`, `key`, `key-type`) operate directly on the DID Document, I don't see how a **DID method** would NOT support them? Those parameters are completely generic.

I could however see that a particular **implementation of a DID resolver** may not support these parameters. E.g. a DID resolver may be able to resolve a DID to its DID Document, but it doesn't have the ability to select services or keys from the DID Document. I agree that probably we want to allow this too.

Does this make sense? Perhaps we mean the same thing anyway?

yancyribbens: > Since all DID methods must be able to resolve a DID to a DID Document, and the first four parameters (service, service-type, key, key-type) operate directly on the DID Document

I agree.  These params don't seem in jeopardy of not being supported by the resolver since these operate on the document returned and is of no consequence to the resolver.

> I could however see that a particular implementation of a DID resolver may not support these parameters. E.g. a DID resolver may be able to resolve a DID to its DID Document, but it doesn't have the ability to select services or keys from the DID Document.

I agree, this shouldn't be of concern to the DID resolver

The parameters which you didn't mention, for example `version-time` would need to operate on a group of DIDs, and therefore be a function of the resolver.  I'm I not understanding how parameters like `version-time` would work?

For parameters such as `version-time`, could these be IF the resolver implements they MUST operate in the same way?

peacekeeper: > I agree, this shouldn't be of concern to the DID resolver

Depends on what we mean by "DID resolver". Is it just the component that returns the DID Document for a given DID, or does it perform the full dereferencing process for a DID URL. We haven't defined this very well yet. This will go into the separate DID Resolution spec.

> I'm I not understanding how parameters like `version-time` would work?

It would return the DID Document for a given DID at a certain point in time (e.g. the one from 2 weeks ago, rather than the current DID Document). This will also be defined in more detail in the DID Resolution spec.

> For parameters such as `version-time`, could these be IF the resolver implements they MUST operate in the same way?

Yes that's the idea.

yancyribbens: > Depends on what we mean by "DID resolver". Is it just the component that returns the DID Document for a given DID, or does it perform the full dereferencing process for a DID URL. We haven't defined this very well yet. This will go into the separate DID Resolution spec.

Great point.  I've been thinking of the DID resolver as the mechanism that dereferences the DID URL.  What happens client side should be a separate concern outside of the DID resolver perhaps?  I do agree that this should be defined better.

> Yes that's the idea.

Great, can we edit this PR so it reads that way?

mwherman2000: > If DID methods can't or won't implement some of the proposed parameters because of a preference for greater decentralization by limiting the heavy lifting done by resolvers, then there should be the option to not implement.

There should also be a away to perform a query against a DID method and DID resolver combination to determine which parameters are supported on a DID method by DID method basis.

common.js
```diff
@@ -84,6 +84,25 @@ var ccg = {
       ],
       status: "CG-DRAFT",
       publisher: "Digital Verification Community Group"
+    },
+    "DRAFT-SPORNY-HASHLINK-02": {
```
msporny: This should just be `HASHLINK`.

peacekeeper: Thanks, fixed in 1a25913e3a340fb623db6c0557102e55609e67a1.

index.html
peacekeeper: @rhiaro and @yancyribbens: Do you think 9ec0477a2e99e4a860c289321bddf41ee2410af9 is a step in the right direction to clarify this?

peacekeeper:
Based on feedback this week, I am creating new PRs that separate 1. Introduction of the matrix parameter concept itself, and 2. Concrete proposed matrix parameters. This way, we can better organize the discussion and accept/change/reject the PRs separately. Please continue here on the new PR:

https://github.com/w3c-ccg/did-spec/pull/189

index.html
peacekeeper: > There should also be a away to perform a query against a DID method and DID resolver combination to determine which parameters are supported on a DID method by DID method basis.

@mwherman2000 As we have discussed in https://github.com/w3c-ccg/did-resolution/issues/25, I agree this is a useful feature that can be supported by DID methods and DID resolvers, however I don't think this is in scope of the DID URL syntax. The proposed grammar always requires a `method-specific-id` as part of the DID URL, therefore there is no way to "perform a query against a DID method" itself using a DID URL.

peacekeeper: Actually @mwherman2000, I'd like to correct myself. https://github.com/w3c-ccg/did-spec/pull/189#discussion_r279237382 just reminded me that in fact the `method-specific-id` can be blank, so with this currently proposed grammar, there could actually be an identifier for the method itself. Happy to explore that further...

PR #189, opened 2019-04-27
Update ABNF to use matrix parameters (new)

### LINKED ISSUE #85: Did Query

### LINKED ISSUE #90: Service endpoint identification matrix parameter

### LINKED ISSUE #170: Inconsistent ABNF and related definitions

### LINKED ISSUE #177: ABNF idchar should include underscore

peacekeeper:
This updates the ABNF and introduces matrix parameters, based on discussions on the mailing list as well as the [DID Spec and DID Resolution Spec Weekly Meetings](https://docs.google.com/document/d/1qYBaXQMUoB86Alquu7WBtWOxsS8SMhp1fioYKEGCabE/) and the resulting [Proposal for Revisions to Decentralized Identifiers Section of DID Spec](https://docs.google.com/document/d/1qnDExIVjU5bYc601qUdLZIi9UAs1ojlHyKnVoz2zlLM/).

**This is a new PR that replaces the earlier https://github.com/w3c-ccg/did-spec/pull/187. In this new one, only the matrix parameter concept itself is introduced, but concrete matrix parameters will be proposed and discussed in separate PRs.**

Separate PRs for concrete matrix parameters: https://github.com/w3c-ccg/did-spec/pull/190, https://github.com/w3c-ccg/did-spec/pull/191, https://github.com/w3c-ccg/did-spec/pull/192, https://github.com/w3c-ccg/did-spec/pull/193, https://github.com/w3c-ccg/did-spec/pull/194, https://github.com/w3c-ccg/did-spec/pull/195, https://github.com/w3c-ccg/did-spec/pull/196

Fixes #85. Fixes #90. Fixes #170. Fixes #177.
Replaces https://github.com/w3c-ccg/did-spec/pull/106. Replaces https://github.com/w3c-ccg/did-spec/pull/168/.
Related to #185.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/189.html" title="Last updated on May 10, 2019, 9:48 AM UTC (1bc70b8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/189/ce1130d...1bc70b8.html" title="Last updated on May 10, 2019, 9:48 AM UTC (1bc70b8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -931,7 +931,7 @@ fragment and MUST conform to the the <code>fragment</code> ABNF rule in
 [[RFC3986]]. A DID fragment MUST be used only as a method-independent
 reference into the DID Document to identify a component of a DID Document
 (e.g. a unique key description). To resolve this reference, the complete
-DID reference including the DID fragment MUST be used as the value of
+DID URL including the DID fragment MUST be used as the value of
 the key for the target component in the DID Document object.
       </p>
 
--- a/index.html
+++ b/index.html
@@ -798,24 +798,24 @@ parameters that MUST operate uniformly across all DID methods. (Method-specific
 parameter names are covered in the following section.)
       </p>
 
-        <table class="simple">
-          <thead>
-            <tr>
-              <th>Parameter Name</th>
-              <th>Description</th>
-            </tr>
-          </thead>
+      <table class="simple">
+        <thead>
+          <tr>
+            <th>Parameter Name</th>
+            <th>Description</th>
+          </tr>
+        </thead>
 
-          <tbody>
-          </tbody>
-        </table>
-        
-        <p>
+        <tbody>
+        </tbody>
+      </table>
+
+      <p>
 The exact processing rules for these parameters are specified in the DID
 Resolution specification.
-        </p>
+      </p>
 
-        <p>
+      <p>
 Note that there may be additional parameters or options that are not part
 of the DID URL but instead passed to a DID resolver “out of band”, i.e.,
 using a resolution protocol or some other mechanism. Such options could
@@ -825,7 +825,7 @@ HTTP headers rather than as part of an HTTP URL. The important distinction
 is that DID parameters that are part of the DID URL specify what resource
 is being identified, whereas DID resolver options that are not part of the
 DID URL control how that resource is dereferenced.
-        </p>
+      </p>
     </section>
 
     <section>
--- a/index.html
+++ b/index.html
@@ -972,7 +972,7 @@ The method name MUST be lowercase.
 
         <li>
 Case sensitivity and normalization of the value of the
-specific-idstring rule in Section <a href="#the-generic-did-scheme">
+specific-idstring rule in Section <a href="#generic-did-syntax">
           </a> MUST be defined by the governing DID method specification.
         </li>
       </ol>
@@ -2057,8 +2057,8 @@ DID Method Schemes
 
       <p>
 A DID method specification MUST define exactly one specific DID
-scheme identified by exactly one method name (the method rule in
-Section <a href="#the-generic-did-scheme"></a>).
+scheme identified by exactly one method name (the <code>method-name</code> rule in
+Section <a href="#generic-did-syntax"></a>).
       </p>
 
       <p>
@@ -2085,7 +2085,7 @@ how to generate the specific-idstring component of a DID. The
 specific-idstring value MUST be able to be generated without the use
 of a centralized registry service. The specific-idstring value SHOULD
 be globally unique by itself. The fully qualified DID as defined by
-the DID rule in Section <a href="#the-generic-did-scheme"></a> MUST
+the DID rule in Section <a href="#generic-did-syntax"></a> MUST
 be globally unique.
       </p>
 
--- a/index.html
+++ b/index.html
@@ -796,7 +796,7 @@ Generic DID Parameter Names
 
       <p>
 DID URL syntax supports a simple, generalized format for parameters based on the
-matrix parameter syntax ([[MatrixURIs]]).
+matrix parameter syntax ([[MATRIX-URIS]]).
 The ABNF above does not specify any parameter names (the <code>param-name</code>
 rule). The following table defines a set of generic DID parameter names for
 parameters that MUST operate uniformly across all DID methods. (Method-specific
--- a/index.html
+++ b/index.html
@@ -802,10 +802,10 @@ The ABNF above does not specify any parameter names (the <code>param-name</code>
 rule).
       </p>
       <p>
-Some generic DID parameter names (e.g. for service selection) are completely
+Some generic DID parameter names (e.g., for service selection) are completely
 independent of any specific DID method and MUST always function the same way
 for all DIDs.
-Others (e.g. for versioning) MAY be supported by
+Others (e.g., for versioning) MAY be supported by
 certain DID methods, but MUST operate uniformly across those DID methods that
 do support them.
       </p>
--- a/index.html
+++ b/index.html
@@ -537,6 +537,7 @@ the controller of the private keys for this identifier.
     "publicKeyPem": "-----BEGIN PUBLIC KEY...END PUBLIC KEY-----\r\n"
   }],
   "service": [{
+    "id": "#service123",
     "type": "ExampleService",
     "serviceEndpoint": "https://example.com/endpoint/8377464"
   }]
@@ -1593,31 +1594,31 @@ Example:
       <pre class="example nohighlight" title="Various service endpoints">
 {
   "service": [{
-    "id": "did:example:123456789abcdefghi;openid",
+    "id": "#openid",
     "type": "OpenIdConnectVersion1.0Service",
     "serviceEndpoint": "https://openid.example.com/"
   }, {
-    "id": "did:example:123456789abcdefghi;vcr",
+    "id": "#vcr",
     "type": "CredentialRepositoryService",
     "serviceEndpoint": "https://repository.example.com/service/8377464"
   }, {
-    "id": "did:example:123456789abcdefghi;xdi",
+    "id": "#xdi",
     "type": "XdiService",
     "serviceEndpoint": "https://xdi.example.com/8377464"
   }, {
-    "id": "did:example:123456789abcdefghi;agent",
+    "id": "#agent",
     "type": "AgentService",
     "serviceEndpoint": "https://agent.example.com/8377464"
   }, {
-    "id": "did:example:123456789abcdefghi;hub",
+    "id": "#hub",
     "type": "HubService",
     "serviceEndpoint": "https://hub.example.com/.identity/did:example:0123456789abcdef/"
   }, {
-    "id": "did:example:123456789abcdefghi;messages",
+    "id": "#messages",
     "type": "MessagingService",
     "serviceEndpoint": "https://example.com/messages/8377464"
   }, {
-    "id": "did:example:123456789abcdefghi;inbox",
+    "id": "#inbox",
     "type": "SocialWebInboxService",
     "serviceEndpoint": "https://social.example.com/83hfh37dj",
     "description": "My public social inbox",
@@ -1626,7 +1627,7 @@ Example:
       "currency": "USD"
     }
   }, {
-    "id": "did:example:123456789abcdefghi;authpush",
+    "id": "#authpush",
     "type": "DidAuthPushModeVersion1",
     "serviceEndpoint": "http://auth.example.com/did:example:123456789abcdefg"
   }]
@@ -1890,7 +1891,7 @@ context above and adding the new property to the <a>DID Document</a>.
   "authentication": [ ... ],
   "service": [<span class="highlight">{
     "@context": "did:example:contexts:987654321",
-    "id": "did:example:123456789abcdefghi;photos",
+    "id": "#photos",
     "type": "PhotoStreamService",
     "serviceEndpoint": "https://example.org/photos/379283"
   }</span>]
@@ -2923,22 +2924,28 @@ A future-facing real-world context is provided below:
   ],
 
   "service": [{
+    "id": "did:example:123456789abcdefghi#oidc",
     "type": "OpenIdConnectVersion1.0Service",
     "serviceEndpoint": "https://openid.example.com/"
   }, {
+    "id": "did:example:123456789abcdefghi#vcStore",
     "type": "CredentialRepositoryService",
     "serviceEndpoint": "https://repository.example.com/service/8377464"
   }, {
+    "id": "did:example:123456789abcdefghi#xdi",
     "type": "XdiService",
     "serviceEndpoint": "https://xdi.example.com/8377464"
   }, {
+    "id": "did:example:123456789abcdefghi#hub",
     "type": "HubService",
     "serviceEndpoint": "https://hub.example.com/.identity/did:example:0123456789abcdef/"
   }, {
+    "id": "did:example:123456789abcdefghi#messaging",
     "type": "MessagingService",
     "serviceEndpoint": "https://example.com/messages/8377464"
   }, {
     "type": "SocialWebInboxService",
+    "id": "did:example:123456789abcdefghi#inbox",
     "serviceEndpoint": "https://social.example.com/83hfh37dj",
     "description": "My public social inbox",
     "spamCost": {
@@ -2947,9 +2954,10 @@ A future-facing real-world context is provided below:
     }
   }, {
     "type": "DidAuthPushModeVersion1",
+    "id": "did:example:123456789abcdefghi#push",
     "serviceEndpoint": "http://auth.example.com/did:example:123456789abcdefghi"
   }, {
-    "id": "did:example:123456789abcdefghi;bops",
+    "id": "did:example:123456789abcdefghi#bops",
     "type": "BopsService",
     "serviceEndpoint": "https://bops.example.com/enterprise/"
   }]
--- a/index.html
+++ b/index.html
@@ -745,7 +745,7 @@ identical to the ABNF rules defined
 in [[RFC3986]].
       </p>
 
-      <p>
+      <p class="note">
 The term <a>DID</a> refers only to the URI
 conforming to the <code>did</code> rule in the ABNF below. A
 DID always identifies the DID subject. The term <a>DID URL</a>,
--- a/index.html
+++ b/index.html
@@ -537,7 +537,7 @@ the controller of the private keys for this identifier.
     "publicKeyPem": "-----BEGIN PUBLIC KEY...END PUBLIC KEY-----\r\n"
   }],
   "service": [{
-    "id": "#service123",
+    "id": "did:example:123456789abcdefghi#service123",
     "type": "ExampleService",
     "serviceEndpoint": "https://example.com/endpoint/8377464"
   }]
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -733,46 +734,178 @@ <h1>
...
-did                = "did:" method ":" specific-idstring
-method             = 1*methodchar
-methodchar         = %x61-7A / DIGIT
-specific-idstring  = idstring *( ":" idstring )
-idstring           = 1*idchar
-idchar             = ALPHA / DIGIT / "." / "-"
+did                = "did:" method-name ":" method-specific-id
+method-name        = 1*method-char
+method-char        = %x61-7A / DIGIT
+method-specific-id = *idchar *( ":" *idchar )
```
mrinalwadhwa: 1. Unless I'm misunderstanding, this rule feels more complicated than it needs to be:
If we change `idchar` to include `:` then a simpler version that would have the same effect:
```
method-specific-id = *idchar
idchar             = ALPHA / DIGIT / "." / "-" / "_" / ":"
```
2. This rule means `method-specific-id` can be blank, is that ok? Maybe we want to require atleast one char:
```
method-specific-id = 1*idchar
idchar             = ALPHA / DIGIT / "." / "-" / "_" / ":"
```

mrinalwadhwa: 1. Unless I'm misunderstanding, this rule feels more complicated than it needs to be:
If we change `idchar` to include `:` then a simpler version that would have the same effect:
```
method-specific-id = *idchar
idchar             = ALPHA / DIGIT / "." / "-" / "_" / ":"
```
2. This rule means `method-specific-id` can be blank, is that ok? Maybe we want to require atleast one char:
```
method-specific-id = 1*idchar
idchar             = ALPHA / DIGIT / "." / "-" / "_" / ":"
```

3. Maybe the intent is to highlight the notion of `:` separated sub-method names in the grammar. If so, then this makes sense. I feel the atleast one char restriction would still makes sense:
```
method-specific-id = 1*idchar *( ":" 1*idchar )
idchar             = ALPHA / DIGIT / "." / "-" / "_"
```

talltree: @mrinalwadhwa You are exactly right that, although the rule could have been simpler, we were wanting to highlight the possibility of using sub-method names in the grammar. Also, there was a specfic request to allow consecutive colons. That also means that the method-specific-id can be blank, however that also satisfies a request for DID method names by themselves to have DID documents, so that was intentional as well.

mrinalwadhwa: got it, thank you for clarifying @talltree

msporny:
> Fixes #85. Fixes #90. Fixes #170. Fixes #177.

I double checked these and agree that those issues would be addressed with this PR.

index.html
```diff
@@ -733,46 +734,178 @@ <h1>
...
+authority components of a DID URI—the <code>path-abempty</code>,
+<code>query</code>, and <code>fragment</code> components are
+identical to the ABNF rules defined
+in [[RFC3986]].
+      </p>
+
+      <p class="note">
+The term <a>DID</a> refers only to the URI
+conforming to the <code>did</code> rule in the ABNF below. A
+DID always identifies the DID subject. The term <a>DID URL</a>,
```
msporny: The DID URL thing makes me uneasy because I feel like developers that use it won't be precise and therefore there will be confusion surrounding the difference between a DID and a DID URL. The confusion is only problematic if the distinction is important, and it may not be that important.

In any case, my concern isn't enough to block the PR... just voicing that I'm concerned about the nuance here and think it may lead to confusion 5 years from now.

index.html
msporny: Hmm, these rules:

```
did                = "did:" method-name ":" method-specific-id
method-specific-id = *idchar *( ":" *idchar )
```

Mean that this is a valid DID:

```
did:v1:
```

I don't think it should be. We should probably change it to this:

```
method-specific-id = 1*idchar *( ":" *idchar )
```

index.html
```diff
@@ -733,46 +734,178 @@ <h1>
...
+      </h2>
+
+      <p>
+DID URL syntax supports a simple, generalized format for parameters based on the
+matrix parameter syntax ([[MATRIX-URIS]]).
+The ABNF above does not specify any parameter names (the <code>param-name</code>
+rule).
+      </p>
+      <p>
+Some generic DID parameter names (e.g. for service selection) are completely
```
msporny: Fix -- "```e.g.,```"

index.html
```diff
@@ -733,46 +734,178 @@ <h1>
...
+independent of any specific DID method and MUST always function the same way
+for all DIDs.
+Others (e.g. for versioning) MAY be supported by
```
msporny: Fix -- "```e.g.,```"

index.html
```diff
@@ -733,46 +734,178 @@ <h1>
...
+
+      <p>
+A DID method specification MAY specify additional method-specific parameter
+names. A method-specific parameter name MUST be prefixed by the method name
+as defined by the <code>method-name</code> rule.
+      </p>
+
+      <p>
+For example, if the method <code>did:foo:</code> defines the parameter bar,
+the parameter name must be <code>foo:bar</code>. An example DID URL using
```
msporny: Suggest not using colon and semantically it's used as a separator elsewhere. I suggest we just do the whole kebab-case thing and do "foo-" instead.

I won't block this PR going in over that, but if we don't change it, I'll probably raise an issue for later on this item.

index.html
```diff
@@ -733,46 +734,178 @@ <h1>
...
+this method and this method-specific parameter would be:
+      </p>
+
+      <p>
+<code>did:foo:21tDAKCERh95uGgKbJNHYp;foo:bar=high</code>
+      </p>
+
+      <p>
+A method-specific parameter name defined by one DID method MAY
```
msporny: oof, this feels like a bad thing to do. If we're going to allow this, there is no reason to prefix it with the method name, don't prefix at all.

index.html
```diff
@@ -733,46 +734,178 @@ <h1>
...
+      </p>
+
+      <p>
+<code>did:example:21tDAKCERh95uGgKbJNHYp;service=agent;foo:bar=high</code>
+      </p>
+
       <p>
-See Sections <a href="#paths"></a> and <a href="#fragments"></a> for the ABNF rules defining DID paths and
-fragments.
+Both DID method namespaces and method-specific parameter
```
msporny: I don't understand the benefit of this. What is the use case? Why are we saying anything about this?

index.html
peacekeeper: On the other hand, I felt like it had been a source of confusion in the last few months to NOT have separate terms here. People said "DID" and sometimes meant only the "naked" / "bare" DID, and sometimes they meant the whole DID+path+query+fragment. We also had the term "DID reference" previously. I used the new terms "DID" and "DID URL" at a [meetup](https://www.meetup.com/Vienna-Digital-Identity-Meetup/events/260079899/) last week and thought it worked well with the (quite technical) audience.

index.html
peacekeeper: @msporny I think I agree with you, however the possibility of having an empty `method-specific-id` has been explicitly discussed. This would make it possible to have a DID Document for the method itself. See https://github.com/w3c-ccg/did-spec/pull/189#pullrequestreview-231512312 and https://github.com/w3c-ccg/did-spec/pull/187#discussion_r278347044.

index.html
peacekeeper: Fixed in 93e9af9004199d5458251e6fb14fd7f892e787dd.

index.html
peacekeeper: Fixed in 93e9af9004199d5458251e6fb14fd7f892e787dd.

index.html
msporny: For DID Documents for the method itself, I think people want "did:v1" not "did:v1:" (note the extra colon at the end). The current ABNF states that "did:v1" is not valid, right? That's the point I'm trying to make.

index.html
msporny: Adding terminology has costs. Sometimes it's clear that those costs are worth it... sometimes it's not clear.

Here's the conversation I'm concerned about:

```
Speaker: So, as we can see did:example:1234 is a URL that can be dereferenced to...
Developer: Wait, that's not a DID URL, that's just a DID... you have to add something like a path  or query or matrix parameter to it for it to be a URL.
Speaker: No, it's a URL, you can dereference it and get a DID Document.
Developer: Then why don't you call it a DID URL? I mean, you can dereference it and get a DID Document... that's the definition of a URL, isn't it!?
```

Do you see the problem we're creating for ourselves? :)

talltree: @msporny Yes, I see the problem clearly, as did other folks attending the DID Spec calls. It was in fact a conscious decision to separate the terms (and the ABNF rules) for DID and DID URL for this reason: a DID document **may** or **may not** be considered a resource as meant the term "URL" (Uniform Resource Locator).

 I know that might sound strange, but let me explain. The analogy we used in our discussion was the resolution of a DNS record in the process of dereferencing the URL for a web page. With such a URL, the resource being located is the web page. It is **not** the DNS record used to look up the IP address of the web server.

So in that sense, when a DID document is used to locate (and describe how to interact with) **another** resource on the Web, the DID document itself is not the resource being located. It is like the DNS record.

However, there are also contexts in which you **do** want to treat a DID document as a locate-able Web resource. For example, if you want to add a fragment to a DID to reference a public key object inside the DID document as a secondary resource. In that case the DID document is the primary resource.

What Markus and I realized is that any reference that needs to treat the DID document as a primary resource needs to add **something** to the DID—turning it into a DID URL. But if you had only a DID with nothing else appended, then the fact that the DID alone **can** be resolved to a DID document does not make that DID document the resource being identified. Rather the DID document is like the DNS record.

Thankfully, this is also consistent with the first statement in the intro paragraph, i.e., that "A DID always identifies the DID subject". If the DID subject is Alice (or Best Buy, or a scooter, or anything else that does not exist on the network), then that DID is clearly a URI that **cannot** function as a URL because the resource it identifies is not locate-able on the Web.

This also thankfully lets us avoid the [HTTP Range 14 ](http://en.wikipedia.org/wiki/HTTPRange-14) issue with DIDs, because the DID by itself always identifies the DID subject and never by itself identifies some other resource describing the DID subject.

index.html
talltree: @msporny Permit me to disagree. I see no issue with requiring the valid DID that identifies a DID method to end in a colon. To me, it's actually more intuitive that way. For example, if I want to refer to the DID scheme by itself, I would write "the did: scheme". If I want to talk about the Veres One DID scheme, I would write "the did:v1: scheme".

index.html
peacekeeper: Just for reference, a more in-depth discussion of this is [here](https://docs.google.com/document/d/1gSUP9DEp7IO8jyNDsVnC-7Ed6PjbMRxl89nGYUoWoeI/).

index.html
peacekeeper: Created issue https://github.com/w3c-ccg/did-spec/issues/198 to keep track of this.

index.html
peacekeeper: Created issue https://github.com/w3c-ccg/did-spec/issues/199 to keep track of this.

index.html
peacekeeper: Created issue https://github.com/w3c-ccg/did-spec/issues/200 to keep track of this and also your other comment below about hierarchical method namespaces.

peacekeeper:
Merging per decision in [2019-05-09 DID Spec and DID Resolution Spec meeting](https://github.com/w3c-ccg/meetings/tree/gh-pages/2019-05-09-did-spec-and-resolution).

PR #191, opened 2019-04-27
Add "service-type" DID URL matrix parameter.

peacekeeper:
This adds one concrete DID URL matrix parameter. See https://github.com/w3c-ccg/did-spec/pull/189.

Description: Identifies a set of services from the DID Document by service type.

Example: `did:example:1234;service-type=agent`

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/191.html" title="Last updated on May 10, 2019, 11:40 AM UTC (4a792ab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/191/fde9370...4a792ab.html" title="Last updated on May 10, 2019, 11:40 AM UTC (4a792ab)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -764,6 +764,15 @@ The following table defines a set of generic DID parameter names:
         </thead>
 
         <tbody>
+          <tr>
+            <td>
+<code>service-type</code>
+            </td>
+            <td>
+Identifies a set of services from the DID Document by service type.
+            </td>
+          </tr>
+
         </tbody>
       </table>
 


index.html
```diff
@@ -825,6 +825,15 @@ <h2>
         </thead>
 
         <tbody>
+          <tr>
+            <td>
+<code>service-type</code>
```
msporny: I don't understand the use case here. Why can't you dereference the DID Document, search for the service type you're looking for, and use that?

index.html
```diff
@@ -825,6 +825,15 @@ <h2>
...
+            </td>
+            <td>
+Identifies a set of services from the DID Document by service type.
```
msporny: How do you use the list of services you get back when this is paired w/ a path?

index.html
peacekeeper: I agree we could just leave it at that. However I think I remember conversations with @dmitrizagidulin that he may see a need for including a service type in the DID URL; some of his examples and considerations are in https://github.com/w3c-ccg/did-spec/issues/90.

Whatever we decide here would probably also apply to https://github.com/w3c-ccg/did-spec/pull/193.

talltree: The specific request for `service-type` came from Daniel Buchner who has said for some months now that the folks working on identity hubs want to be able to request an array of all the services of that type (and that it will be typical for a DID controller to have more than one).

So while clearly the calling application could just request the DID document and then parse it itself, this matrix parameter would enable a DID URL to specify that specific service type, and thus for a resolver to follow the same algorithm for constructing a concrete URL as it would for a `service` parameter, only it would return an array of concrete URLs, not just one.

index.html
peacekeeper: The path/query/fragment of the DID URL would be applied to ALL service endpoints in the resulting set.

peacekeeper:
Quick note that in the [2019-05-30 DID Spec and DID Resolution Spec meeting](https://github.com/w3c-ccg/meetings/tree/gh-pages/2019-05-30-did-spec-and-resolution) we discussed this PR, and specifically the question whether we need matrix parameters for service selection, versioning, etc., or whether such functionality should be supported in a different layer (i.e. DID Resolution protocol). In the meeting, there was support for this matrix parameter, but we'll continue to discuss a bit more before merging.

brentzundel:

peacekeeper:
Closing. This PR has been re-created in the DIDWG repo: https://github.com/w3c/did-spec/pull/62

Issue #64, opened 2018-03-10
DID Document versioning


Drabiv:
I was thinking on how to make audit of DID Document updates as easy and as transparent as possible. 
How resolver of DID makes sure that it has that latest version of DID Document? If resolver needs to get the whole history of DID Documents for DID how it makes sure that it has a whole history without any omissions?
I realise that this is DLT and DID method specific, but still are there any recommendations or explanation of the best way it should be done or about MUST requirements for this. Are any explanations of this for implemented DID Method specifications availalble?

Also, to make audit of DDoc history as transparent and as simple as possible I propose to add "updateNumber" property to DID Documents. On every DDoc update the value of updateNumber MUST grow by 1. Two or more DDocs with the same updateNumber, but with different content are NOT ALLOWED. This property would allow to have a clear audit trail of changes. It will allow to know from one glance how many changes happened with that DDoc.

msporny:
We should raise this issue with the CCG. The proposal is to make versioning a responsibility of the DID Method where the version would go on the transaction/operation that is sent to or received from the ledger. That is, the version is metadata around the DID Document.

kimdhamilton:
tl;dr I think we should move this issue to a CCG repo; most likely the DID method registry. 

Right, at least informally, we've been approaching this as a responsibility of the DID method, in terms of the requirements it imposes on the resolver to ensure the latest version is obtained. 

Regarding update numbers: for BTCR, we've specified updates as spending the output of an existing transaction; so in order to find the latest, all we care about is finding the end of the transaction trail and ensuring signatures (in continuation document/s) are consistent. In this case, the need to specify an `updateNumber` property in the continuation doc might be more of a bookkeeping burden than useful for disambiguation (but the resolver could calculate it if the property were desired...)

That said, in BTCR we've informally discussed the ability to update the continuation documents outside of the transactions (if the continuation document is not in an immutable ledger). In this case, an update number might be interesting, but hard to discover conflicts and enforce.

This brings up several points:
- If it is a responsibility of the method specs, this needs to be made explicit. It's clear that method specs must describe how to update the DID doc, but it's not clear they always address finding "the latest" (BTCR does but BTCR updates are entirely tied to the  ...)
- We've discussed in the past the idea of getting a DID document at a specific point in time, if the DID method and ledger allow it. I'm still interested in keeping this scenario in mind (for longer-lived claims)
[...]

PR #194, opened 2019-04-27
Add "version-id" and "version-time" DID URL matrix parameters.

peacekeeper:
This adds two concrete DID URL matrix parameters. See https://github.com/w3c-ccg/did-spec/pull/189.

Description: Identifies a specific version of a DID Document to be resolved.

Example: `did:example:1234;version-id=3` or `did:example:1234;version-time=1556363673`

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/194.html" title="Last updated on Aug 2, 2019, 6:45 PM UTC (3f281a0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/194/87676ee...3f281a0.html" title="Last updated on Aug 2, 2019, 6:45 PM UTC (3f281a0)">Diff</a>


msporny:
This feels like something that should be in the DID Resolution protocol, not in the DID URL itself.

I don't understand the use case that requires this feature? When do you need to do something like this?

peacekeeper:
@msporny 
> I don't understand the use case that requires this feature? When do you need to do something like this?

The idea is that you want to be able to know what a certain public key or service endpoint was at a specific point in time. E.g. you want to validate a longer-lived VC that was signed by an issuer 2 years ago, but the issuer's keys in their DID Document have been rotated since then. Or you want a stable reference to a service endpoint as it was in some previous version of a DID Document rather than the current service endpoint.

So you could do things like this:

`did:btcr:xz35-jzv2-qqs2-9wjt;version-time=1557005485#keys-1`

Or

`did:btcr:xz35-jzv2-qqs2-9wjt;version-time=1557005485;service=myprofile/myphoto.jpg`

> This feels like something that should be in the DID Resolution protocol, not in the DID URL itself.

I agree this parameter could be passed to a DID Resolver separately from the DID URL, as an out-of-band option, but it can also make sense to have it as part of the DID URL in order to essentially make all historic DID Document information part of the URL's addressing space. I think anything that affects WHAT is being identified should be part of the DID URL.

See also https://github.com/w3c-ccg/did-spec/issues/64 and https://github.com/w3c-ccg/did-resolution/issues/12.

msporny:
> I agree this parameter could be passed to a DID Resolver separately from the DID URL, as an out-of-band option, but it can also make sense to have it as part of the DID URL in order to essentially make all historic DID Document information part of the URL's addressing space.

Ah, ok, understand the use case now (and it's what I thought it was).

I think this is a layer violation in the architecture and, in general, that the query protocol shouldn't be mixed into the matrix parameters. 

To put this another way, we can solve this problem higher up on the stack (in the DID Resolution protocol, which may get very complex over the course of 10 years... which means we shouldn't set a precedent where the matrix parameters carry every possible thing you can pass to the DID Resolution process).

I'd rather we are very conservative w/ features w/ the 1.0 release, put optional things in the DID Resolution layer, and only move them down in the stack if it's clear that there is a pain point here.

peacekeeper:
@msporny I understand that perspective, and I agree there will be many parameters/options on the DID Resolution layer that will never be expressed with matrix parameters or otherwise in the DID URL. E.g. caching behavior, this should definitely be controlled higher up on the stack. I can also see us treating versioning functionality the same way.

But for the sake of argument, let me try to make a point why it could still make sense to include versioning information in the DID URL: You would basically make previous versions of DID Documents addressable, and you could reference all those previous versions. For example, in a Linked Data Proof, you could explicitly reference a key from a specific version of a DID Document:

```
{
  "@context": "https://w3id.org/identity/v1",
  "title": "Hello World!",
  "proof": {
    "type": "RsaSignature2018",
    "creator": "did:btcr:xz35-jzv2-qqs2-9wjt;version-time=1557005485#keys-1",
    "proofValue": "eyJ0eXAiOiJK...gFWFOEjXk"
  }
}
```

I've been thinking of weblogs as an analogy. If you open the front page URL of a blog in your browser, you see the latest article. But all historic articles are also available on the blog. If you want to retrieve those historic articles, you don't just open the same front page URL and pass in a special HTTP header for versioning. That would actually violate web architecture, since the same URL can't identify more than one resource. So instead, all historic articles on that blog have their own URLs (permalinks). With versioning matrix parameters, we'd basically be creating permalinks for all historic DID Documents (and their keys, service endpoints, content references, etc.).

Maybe the analogy is flawed, but this was my thinking; like I said I'd also be happy to specify this as parameters on the DID Resolution layer, similar to caching and others.

We could also decide to not list ANY matrix parameters in the main DID spec and move them all into the DID Resolution spec.

peacekeeper:
> you don't just open the same front page URL and pass in a special HTTP header for versioning. 

Actually, it seems there is an HTTP header for that: Accept-Datetime ([RFC7089](https://tools.ietf.org/html/rfc7089)). It says:

_The "Accept-Datetime" request header is transmitted by a user agent to indicate it wants to access a past state of an Original Resource._

So I guess that supports your argument.

cboscolo:
If we do end up enabling a way to resolve DID Documents via a version id, we will need to be careful with how public keys are referenced when used within protocols such as DID Auth.

We don't want to allow an attack where someone can use an old compromised key by referencing a DID Document by the version for which the key would work.

PR #204, opened 2019-05-30
Implement browser support.

mattcollier:


dmitrizagidulin:
Oh this is awesome! What's the usage / how do I consume the webpack'd contexts?

peacekeeper:
@mattcollier can you add some additional info what this does? E.g. does this make it easier to include the context JSON files in web applications?

davidlehn:
The packaging is for https://www.npmjs.com/package/did-context.  The updates make it so the package can be used from both Node.js and now for frontend use through bundlers like webpack.  A few solutions were looked at and we ended up with the webpack/rollup thing to inline contexts for frontend use.  The build is a bit convoluted at the moment but will hopefully be simplified with future webpack ESM features.

davidlehn:
NPM packaging updated and moved to: https://github.com/digitalbazaar/did-context.  This issue can be closed and current packaging files can be removed with https://github.com/w3c-ccg/did-spec/pull/232.

rhiaro:
Closing, superceded by #232

Issue #205, opened 2019-05-31
Add "DID controller" to "Terminology" section.


peacekeeper:
Also check if "DID controller" should be used in additional places, e.g. see https://github.com/w3c-ccg/did-spec/pull/180/files#r283199888

rhiaro:
DID Controller is now in the terminology section, but leaving this open until we do a sweep to make sure we refer to it appropriately throughout the spec.

jandrieu:
Closing because we've moved it to the DID WG did-spec repo.

PR #242, opened 2019-08-01
Add `controller` pattern example (also replaces Guardianship).

dlongley:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/pull/242.html" title="Last updated on Aug 1, 2019, 8:20 PM UTC (f9e9959)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/242/7b2cf89...f9e9959.html" title="Last updated on Aug 1, 2019, 8:20 PM UTC (f9e9959)">Diff</a>

PR #255, opened 2019-08-08
Fix inconsistent use of term 'DID subject'

dhh1128:
Signed-off-by: Daniel Hardman <daniel.hardman@gmail.com>

***
<a href="https://pr-preview.s3.amazonaws.com/dhh1128/did-spec/pull/255.html" title="Last updated on Aug 8, 2019, 9:38 AM UTC (8d99ae0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-ccg/did-spec/255/bd4f2fa...dhh1128:8d99ae0.html" title="Last updated on Aug 8, 2019, 9:38 AM UTC (8d99ae0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1333,14 +1333,14 @@ Authentication
       </h2>
 
       <p>
-Authentication is the mechanism by which a <a href="#did-subject">DID subject</a> can
-cryptographically prove that they are associated with a DID.
+Authentication is the mechanism by which the controller(s) of a DID can
+cryptographically prove that they are associated with that DID.
 See Section <a href="#binding-of-identity"></a>. Note
-that Authentication is separate from Authorization because the subject
+that Authentication is separate from Authorization because the controllers
 may wish to enable others to update their DID Document (for
 example, to assist with key recovery as discussed in Section
 <a href="#key-revocation-and-recovery"></a>) without enabling them to prove
-control (and thus be able to impersonate the subject).
+control (and thus be able to impersonate the controllers).
       </p>
 
       <p>


dhh1128:
The use of the term "DID subject" in section 5.2 DID Subject, section 5.4 Authentication, and 9.3 Binding of Identity is inconsistent:

>5.2: "The DID subject is the identifier that the DID Document is about, i.e., it is the DID described by DID Document." --> subject = the DID

>5.4: "Authentication is the mechanism by which a DID subject can cryptographically prove that they are associated with a DID...the subject may wish to enable others to update their DID Document...thus be able to impersonate the subject" --> subject = person who owns the DID

I believe that 5.4 could be reworded to use the word "controller", and that this would resolve much of the inconsistency, so that's what I've done in this PR.

However, I find section 9.3.1 to be confusing, both before and after this PR--largely because I don't understand what entity is proving control. DIDs can't prove control; they're inert strings. A controller could prove control of a DID+DID Doc combination--but this requires use of a public key, which is described in section 9.3.2. So I'm left scratching my head a bit. This may mean that my PR is based on faulty understanding--but if so, it's a symptom that the language needs to be clarified, because I have read this carefully, several times, and I have a moderately deep understanding of the problem space, and I can't figure it out.

@peacekeeper @msporny @talltree

index.html
```diff
@@ -1333,14 +1333,14 @@ <h2>
       </h2>
 
       <p>
-Authentication is the mechanism by which a <a href="#did-subject">DID subject</a> can
-cryptographically prove that they are associated with a DID.
+Authentication is the mechanism by which the controller(s) of a DID can
```
peacekeeper: We should probably use the term "DID controller" here, since just last week we merged https://github.com/w3c-ccg/did-spec/pull/242/ which uses this term and adds an example.

And perhaps we can still add that term to the "Terminology" section to fix https://github.com/w3c-ccg/did-spec/issues/205.

rhiaro:
I agree that 5.2:

> The DID subject is the identifier that the DID Document is about, i.e., it is the DID described by DID Document.

is confusing because the DID subject *isn't* the *identifier* is it? The attributes in the DID Document are not about that string, but about something identified by that string. I hesitate to use the word 'entity', but the intro (which I think was going to get a rework at some point and may not be authoritative text) says

> the entity the DID identifies (aka the DID subject)

Unless this is a completely wrong interpretation of what DID Subject is, we could make 5.2 consistent with the intro, to say something like:

> The DID Subject is the entity that the DID Document is about, i.e., it is the entity identified by the DID and described by the DID Document.

or hedging a bit:

> The DID Subject is the what the DID Document is about, i.e., it is something which the DID identifies and is described by the DID Document.

And we could add the definition of DID Subject to Terminology as well.

peacekeeper:
>  the DID subject _isn't_ the _identifier_ is it? 

Fully agree with everything @rhiaro said, looks like we need to fix 5.2.

We spent several weeks discussing this, including the question what is it that the DID (URL) identifies. Just for reference, [here](https://docs.google.com/document/d/1gSUP9DEp7IO8jyNDsVnC-7Ed6PjbMRxl89nGYUoWoeI/) is a summary of what we discussed back then (no need to repeat the details now).

PR #270, opened 2019-08-19
Add publicKeyHex as a valid publicKey format

### LINKED ISSUE #152: publicKeyJwk, publicKeyHex, publicKeyBase64, publicKeyBase58 missing from context.
Per the spec, https://w3c-ccg.github.io/did-spec/#public-keys

I was expecting to see publicKeyPem, publicKeyJwk, publicKeyHex, publicKeyBase64 to be present here:

https://w3id.org/did/v1, which resolves to https://w3c-ccg.github.io/did-spec/contexts/did-v1.jsonld

This causes errors when signing DID Documents which contain publicKeys with properties missing from the context, for example:

```
const https = require("https");
const jsonldSignatures = require("jsonld-signatures");

const get = url => {
  return new Promise((resolve, reject) => {
    https
      .get(url, resp => {
        let data = "";
        resp.on("data", chunk => {
          data += chunk;
        });
        resp.on("end", () => {
          resolve(JSON.parse(data));
        });
      })
      .on("error", err => {
        reject(err);
      });
  });
};

const { EcdsaKoblitzSignature2016 } = jsonldSignatures.suites;
const { PublicKeyProofPurpose } = jsonldSignatures.purposes;

(async () => {
  const didDocument = {
    "@context": "https://w3id.org/did/v1",
    id: "did:example:123",
    publicKey: [
      {
        id:
          "did:example:123#kid=1b96181835f46197512e826e1e5c06d2ce45a3518ee9d975938aab899de4c933",
        type: "EcdsaKoblitzSignature2016",
        owner: "did:example:123",
        publicKeyHex:
          "02234be9bcdf041f7530979b8b88b7dc62dd505a75883c8211f3a8250534f96dc0"
      }
    ],
    authentication: [
      {
        publicKey:
          "did:example:123#kid=1b96181835f46197512e826e1e5c06d2ce45a3518ee9d975938aab899de4c933",
        type: "EcdsaKoblitzSignature2016"
      }
    ]
  };

  const didv1 = await get(
    "https://w3c-ccg.github.io/did-spec/contexts/did-v1.jsonld"
  );
  const testLoader = url => {
    // console.log(url);

    if (url === "https://w3id.org/did/v1") {
      return {
        contextUrl: null,
        document: didv1,
        documentUrl: url
      };
    }

    if (
      url ===
      "ecdsa-koblitz-pubkey:02234be9bcdf041f7530979b8b88b7dc62dd505a75883c8211f3a8250534f96dc0"
    ) {
      return {
        contextUrl: null,
        document: {
          "@context": "https://w3id.org/security/v2",
          id:
            "ecdsa-koblitz-pubkey:02234be9bcdf041f7530979b8b88b7dc62dd505a75883c8211f3a8250534f96dc0",
          type: "CryptographicKey",
          owner: "https://example.com/i/alice",
          publicKeyWif: "16agVsYKpbHL4H4e6NSm4yDkpRN4w84E2Q"
        },
        documentUrl: url
      };
    }
  };

  const controller = {
    "@context": "https://w3id.org/security/v1",
    id: "https://example.com/i/alice",
    publicKey: [
      {
        "@context": "https://w3id.org/security/v1",
        id:
          "ecdsa-koblitz-pubkey:02234be9bcdf041f7530979b8b88b7dc62dd505a75883c8211f3a8250534f96dc0",
        type: "CryptographicKey",
        owner: "https://example.com/i/alice",
        publicKeyWif: "16agVsYKpbHL4H4e6NSm4yDkpRN4w84E2Q"
      }
    ],
    assertionMethod: [
      "ecdsa-koblitz-pubkey:02234be9bcdf041f7530979b8b88b7dc62dd505a75883c8211f3a8250534f96dc0"
    ]
  };

  const signed = await jsonldSignatures.sign(didDocument, {
    documentLoader: testLoader,
    suite: new EcdsaKoblitzSignature2016({
      privateKeyWif: "KzVDttSotdDwhfBBiaU4vPKK3yeUmvfLcWbBPHdWy2k3wNYs2oPE",
      creator: `ecdsa-koblitz-pubkey:02234be9bcdf041f7530979b8b88b7dc62dd505a75883c8211f3a8250534f96dc0`
    }),
    purpose: new PublicKeyProofPurpose()
  });
  const result = await jsonldSignatures.verify(signed, {
    documentLoader: testLoader,
    suite: new EcdsaKoblitzSignature2016({
      creator: `ecdsa-koblitz-pubkey:02234be9bcdf041f7530979b8b88b7dc62dd505a75883c8211f3a8250534f96dc0`,
      date: "2017-03-25T22:01:04Z"
    }),
    purpose: new PublicKeyProofPurpose({
      controller
    })
  });

  console.log("verified? ", result);
})();
```

`The property "publicKeyHex" in the input was not defined in the context.`

gjgd:
fixes #152 

Added publicKeyHex as a valid publicKey format

The fact that `publicKeyHex` is missing from the did spec as a valid `publicKey` format is preventing us from adding [did:elem](https://github.com/decentralized-identity/element) to the [universal resolver](https://uniresolver.io/)

Not sure if can just edit the v0.11 file or if I should create a v0.12, let me know!

### DIFF
--- a/contexts/did-v0.12.jsonld
+++ b/contexts/did-v0.12.jsonld
@@ -50,6 +50,7 @@
     "publicKey": {"@id": "sec:publicKey", "@type": "@id", "@container": "@set"},
     "publicKeyBase58": "sec:publicKeyBase58",
     "publicKeyPem": "sec:publicKeyPem",
+    "publicKeyHex": "sec:publicKeyHex",
     "revoked": {"@id": "sec:revoked", "@type": "xsd:dateTime"},
     "service": {"@id": "didv:service", "@type": "@id", "@container": "@set"},
     "serviceEndpoint": {"@id": "didv:serviceEndpoint", "@type": "@id"},
--- a/contexts/did-v0.12.jsonld
+++ b/contexts/did-v0.12.jsonld
@@ -1,6 +1,6 @@
 {
   "@context": {
-    "@version": 1.2,
+    "@version": 1.1,
     "id": "@id",
     "type": "@type",
 
--- a/contexts/did-v0.12.jsonld
+++ b/contexts/did-v0.12.jsonld
@@ -1,6 +1,6 @@
 {
   "@context": {
-    "@version": 1.1,
+    "@version": 1.2,
     "id": "@id",
     "type": "@type",
 
--- a/contexts/did-v0.11.jsonld
+++ b/contexts/did-v0.11.jsonld
@@ -50,6 +50,7 @@
     "publicKey": {"@id": "sec:publicKey", "@type": "@id", "@container": "@set"},
     "publicKeyBase58": "sec:publicKeyBase58",
     "publicKeyPem": "sec:publicKeyPem",
+    "publicKeyHex": "sec:publicKeyHex",
     "revoked": {"@id": "sec:revoked", "@type": "xsd:dateTime"},
     "service": {"@id": "didv:service", "@type": "@id", "@container": "@set"},
     "serviceEndpoint": {"@id": "didv:serviceEndpoint", "@type": "@id"},


dmitrizagidulin:
@gjgd Ok, so, we're about to submit a PR (parallel to this one) that explains the context versioning mechanism (in a README in the `contexts/` folder), which seems to have buy-in from at least some of the devs maintaining the contexts.

With that in mind, I recommend - make a v0.12 version (in addition to the v0.11), and add the new term to /that/ one.

gjgd:
Thanks @dmitrizagidulin , I put it in a separate file

contexts/did-v0.12.jsonld
```diff
@@ -1,6 +1,6 @@
 {
   "@context": {
-    "@version": 1.1,
+    "@version": 1.2,
```
dmitrizagidulin: Heh, sorry for the confusion - the `@version` property is for the JSON-LD version (so, only 1.1), not the context version. (Kind of like how currently HTTP calls in your browser are made to HTTP 1.1 etc)

contexts/did-v0.12.jsonld
```diff
@@ -0,0 +1,59 @@
...
+    "keyAgreement": {"@id": "sec:keyAgreementMethod", "@type": "@id", "@container": "@set"},
+    "nonce": "sec:nonce",
+    "owner": {"@id": "sec:owner", "@type": "@id"},
+    "proof": {"@id": "sec:proof", "@type": "@id", "@container": "@graph"},
+    "proofPurpose": {"@id": "sec:proofPurpose", "@type": "@vocab"},
+    "proofValue": "sec:proofValue",
+    "publicKey": {"@id": "sec:publicKey", "@type": "@id", "@container": "@set"},
+    "publicKeyBase58": "sec:publicKeyBase58",
+    "publicKeyPem": "sec:publicKeyPem",
+    "publicKeyHex": "sec:publicKeyHex",
```
msporny: We really need to move to publicKeyMultibase ... because that supports base16 (hex), base58, base64url, etc.

Can we do that instead?

msporny: We should probably also switch to Multicodec encoding of public keys... we've been needing to do this for a very long time.

dmitrizagidulin: Agreed on both counts (multibase and multicodec). Although I wonder if it might cause less confusion to at least have a transition period where we support both a general `publicKeyMultibase` and the specific ones `publicKeyBase58` etc, side by side.

gjgd: We could add `publicKeyMultibase` but then it wouldn't be consistent to keep `publicKeyBase58`. The problem is removing `publicKeyBase58` would cause retro-compatibility issue because it is already used by did methods (did:sov)

gjgd: Yes, I agree we could support them all side by side for now

dmitrizagidulin: @msporny - are you ok with supporting both?

msporny: > The problem is removing publicKeyBase58 would cause retro-compatibility issue because it is already used by did methods (did:sov)

Yes, but remember, if we bump to did-v0.12, we can drop whatever we want to and not affect the people using did-v0.11. We can drastically change the JSON-LD contexts because they are versioned. 

Now, the problem that creates is that Sovrin won't be able to understand publicKeyMulticodec... but it should bail anyway if it sees a did-v0.12 context.

So, we'd drop publicKeyBase58 when moving over to publicKeyMulticodec... and as you can probably tell, I'm making this up as I go along and haven't had enough time to really understand the impacts of this PR (but have been thinking about making this switch for 8+ months, and maybe this is the PR that pushes the issue and gets us to make the call).

msporny: > @msporny - are you ok with supporting both?

No, let's try to do the right thing and get the community on board and not some funky pseudo-compatability thing. I don't see much of an upside of multiple ways to represent keys... we should see if we can multicodec the RSA public key and JWK stuff as well.

If we're going to do this, let's do it the right way.

dmitrizagidulin: @msporny I agree that bumping the version on the spec would help shield the existing implementations from breaking.

On the other hand, the benefits of supporting the original specific terms (publicKeyBase58 and so on) in addition to the new, for at least several versions while we build consensus, outweighs the downsides.

dmitrizagidulin: @msporny ok, makes sense. Can we add this issue to tomorrow call's agenda, by any chance?

msporny: > On the other hand, the benefits of supporting the original specific terms (publicKeyBase58 and so on) in addition to the new, for at least several versions while we build consensus, outweighs the downsides.

I strongly disagree... because it forces everyone in the ecosystem to write compatability if/then statements in every library that has to detect publicKeyBase58, publicKeyJwk, publicKeyHex, publicKeyMulticodec and so on. It would be better to not have to write that compatibility code (in security libraries), which further complicates implementations resulting in more brittle security code.

msporny: > @msporny ok, makes sense. Can we add this issue to tomorrow call's agenda, by any chance?

That's a call for the CCG Chairs (@ChristopherA, @jandrieu, @kimdhamilton) or the DID Task Force Chairs (@talltree, @peacekeeper)...

gjgd: Can we merge this PR as is, and create an issue for `publicKeyMulticodec` that can be discussed and solved a separate PR?

msporny: No, we shouldn't merge. This should be a CCG decision... it has far too much of an impact on DID Method implementers to merge w/o consulting w/ the community first.

OR13: I don't see why we would not add support for the key types listed in the spec:
https://w3c-ccg.github.io/did-spec/#public-keys

I agree, that some are better than others (Personally I'm a fan of publicKeyPem and publicKeyJwk). 

We should either remove the key types from the spec or support them. Saying the the community uses a key type, but then having it not be supported seems wrong.

I agree that the CCG should make the decision here.

peacekeeper: I'm with @dmitrizagidulin and @OR13 on this.. I think `publicKeyMulticodec` is great. But considering that we just released the [Community Group Final Report](https://w3c-ccg.github.io/did-spec/CGFR/2019-08-10/), we should have a context (either the current v0.11 or a new v0.12, doesn't matter so much I think) that contains all terms used in that released spec, including all the example public key properties (see [§ 5.3](https://w3c-ccg.github.io/did-spec/CGFR/2019-08-10/#public-keys), this was last changed in https://github.com/w3c-ccg/did-spec/commit/8dbfe62a3bb6841959c8af349b29b749fdd19ec2).

After that we could immediately start working on a new context 0.13 and have the discussion about removing everything except `publicKeyMultibase` / `publicKeyMulticodec`.

OR13:
This PR is semi-blocking the DIF interop project from demonstrating interoperability between DDOs with publicKeyHex and publicKeyJwk. 

Per the spec: https://w3c-ccg.github.io/did-spec/#public-keys

```
The following is a non-exhaustive list of public key properties used by the community: publicKeyPem, publicKeyJwk, publicKeyHex, publicKeyBase64, publicKeyBase58, publicKeyMultibase, ethereumAddress.
```
There does not appear to be a single commited / hosted context that supports all of these, but there are context that support `publicKeyBase58`.

Additionally `https://w3id.org/did/v0.11` redirects to `https://w3c-ccg.github.io/did-spec/contexts/did-v0.11.jsonld`, is not a hashlink, and provides no assurance that v0.11 context is immutable.

It seems likely that without a W3 hosted context that supports community key types, community members are likely to host their own contexts, is that a good thing or a bad thing?

I have been assuming that the ideal scenario is that all DID Methods would try their best to support the same latest context. 

Maybe it is preferred that each DID Method host their own context which describes the key types and custom properties present in their DDO?

It feels like a security risk to point to a context you don't control, am I wrong to assume that?

OR13:
It seems like community hosting of contexts won't work... and this PR is not sufficient to fix this issue because:

```
[
      {
        "@id": "did:elem:eURSFEEv6J7s3TJ-jhT_ZS4uGRyCDbwc347EWlqpNgw",
        "https://w3id.org/security#publicKey": [
          {
            "@id": "did:elem:eURSFEEv6J7s3TJ-jhT_ZS4uGRyCDbwc347EWlqpNgw#key-0"
          },
          {
            "@id": "did:elem:eURSFEEv6J7s3TJ-jhT_ZS4uGRyCDbwc347EWlqpNgw#key-JUvpllMEYUZ2joO59UNui_XYDqxVqiFLLAJ8klWuPBw"
          }
        ]
      },
      {
        "@id": "did:elem:eURSFEEv6J7s3TJ-jhT_ZS4uGRyCDbwc347EWlqpNgw#key-0",
        "@type": [
          "https://w3id.org/security#EcdsaSecp256k1VerificationKey2019"
        ],
        "https://w3id.org/security#publicKeyHex": [
          {
            "@value": "027560af3387d375e3342a6968179ef3c6d04f5d33b2b611cf326d4708badd7770"
          }
        ]
      },
      {
        "@id": "did:elem:eURSFEEv6J7s3TJ-jhT_ZS4uGRyCDbwc347EWlqpNgw#key-JUvpllMEYUZ2joO59UNui_XYDqxVqiFLLAJ8klWuPBw",
        "@type": [
          "https://w3id.org/security#EcdsaSecp256k1VerificationKey2019"
        ],
        "https://w3id.org/security#publicKeyJwk": [
          {
            "@id": "_:b0"
          }
        ]
      }
    ]
```

https://web-payments.org/vocabs/security#EcdsaSecp256k1VerificationKey2019
https://w3id.org/security#publicKeyJwk

These links won't point to anything, but does that really matter?

If it does, then there seems to be a tight coupling between valid context updates and the html content of https://w3id.org...

Based on https://github.com/w3c-ccg/did-spec/pull/272

It seems like we may want to add some additional criteria to the versioning requirements, namely that there not be any broken links in the contexts.

OR13:
After a few more days of pondering, I created a self hosted solution for this.

There appears to be a tight coupling between contexts and their documentation.

When we update a context, there should be no broken links, new properties should be documented! 

I don't think its possible to do that with the hosting setup that exists currently, or it will require several PRs across a number of repos, and I suspect that will lead to no actual documentation for contexts, as is the case currently. For example, where is the documentation for:

https://w3id.org/security/v2 ?

None of these properties are defined here: https://w3id.org/security#, these docs were last updated in 2016.

This PR: https://github.com/w3c-ccg/did-spec/pull/207

Added properties to the context that are not documented, and some of these properties are actually not usable without #270  which adds support for the key types used by `EcdsaSecp256k1Signature2019`...

As the DID Spec moves along, I would encourage us to host context and documentation in the same place, so that contributors can fix issues like this easier in the future. 

If the repo example I have linked is useful for that purpose, feel free to copy it:

https://github.com/transmute-industries/context/issues/1

Just to be super clear, I want to use the W3 hosted did spec without extending it, and I want to contribute to it. 

Contribution should not result in broken links, and should not need to be spread across multiple repos, because that will cause broken links, and slow down contribution.

Should I open issues for undocumented context properties in the did spec and cross link to these repo's?:

https://github.com/perma-id/w3id.org/blob/master/security/.htaccess
https://github.com/web-payments/web-payments.org

I'm not trying to be a pest, clean documentation will help us attract more contributors, and will make all of our lives easier. I'm happy to help accomplish this however I can.

msporny:
@OR13 you're catching us at a really heavy time of conferences and travel. Almost everyone in the community is traveling to/from RWoT9, APConf2019, W3C TPAC 2019, and a variety of other things. There are answers to some of your questions and plans for others, but I need to catch a 5am flight and need to go to sleep. Others are in the same position. Can you request some time to talk about this on a CCG call in early October? Self-hosting the solution (or ideally just overloading the JSON-LD document loader) should work for you for the time being.

OR13:
@msporny thanks! I will request some time on the call :)

brentzundel:
The artifacts in the DID WG repository (including the context document) share commit history with this repo, so it should be possible to raise this PR [against that repo](https://github.com/w3c/did-spec/pulls).

peacekeeper:
@gjgd as discussed at IIW29, do you think you could re-create this PR in the DIDWG repo: https://github.com/w3c/did-spec/pulls and then close it here?

gjgd:
Yes, I'm not sure we want to merge it as is anymore but it's done @peacekeeper 
https://github.com/w3c/did-spec/pull/63

peacekeeper:
Thanks, we can continue to discuss at https://github.com/w3c/did-spec/pull/63. Closing.

PR #272, opened 2019-08-19
Add a contexts/README.md to explain the DID context versioning scheme.

dmitrizagidulin:
Add an explanation and changelog for the DID `@context` versioning scheme.

(Related to PR #270 and others.)

### DIFF
--- a/contexts/README.md
+++ b/contexts/README.md
@@ -1,0 +1,43 @@
+# DID Contexts
+
+## Background
+
+JSON-LD Context versioning (and schema versioning in general) is a complex 
+topic, involving the need to balance security and developer convenience.
+
+On one end of the spectrum, some use cases require cryptographically bound and 
+immutable contexts, and will need to use a contest-addressable URL mechanism,
+such as the in-progress 
+[Hashlink](https://tools.ietf.org/html/draft-sporny-hashlink-03) spec.
+
+On the opposite end of the spectrum, developers that can tolerate the fact that
+the DID context is still changing (in the Credentials Community Group and the 
+forthcoming DID Working Group), may want a way to just say "I want to use the
+latest in-progress `@context`", without constraining the versions.
+
+And between the two extremes, some developers may want to use strongly versioned
+contexts (without necessarily cryptographically binding assurance).
+
+## DID Context Versions
+
+The current system for DID context versioning is:
+
+1. Use `https://www.w3.org/2019/did/v1` as an alias to the latest version of the
+spec. Once the DID Working Group work concludes, the final v1 version will be
+frozen and made immutable.
+2. Use intermediate pre-v1 versioned URLs such as `https://w3id.org/did/v0.11`
+if you need to refer to a specific context version.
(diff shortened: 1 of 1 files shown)


contexts/README.md
```diff
@@ -0,0 +1,43 @@
+# DID Contexts
+
+## Background
+
+JSON-LD Context versioning (and schema versioning in general) is a complex 
+topic, involving the need to balance security and developer convenience.
```
TallTed: `security and developer`
-> `security with developer`

contexts/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+On one end of the spectrum, some use cases require cryptographically bound and 
+immutable contexts, and will need to use a contest-addressable URL mechanism,
```
TallTed: `contest-addressable`
-> `context-addressable`

dlongley:
```suggestion
immutable contexts, and will need to use a content-addressable URL mechanism,
```

dlongley: Note: You don't necessarily "need" to use a content-addressable URL mechanism... your software could instead force the resolution of a given URL to some hard-coded content (which is what will be expected for those using the v1 DID context once its done).

contexts/README.md
dmitrizagidulin: Thanks!

contexts/README.md
dmitrizagidulin: Good point, will add a note to that effect.

contexts/README.md
```diff
@@ -6,9 +6,10 @@ JSON-LD Context versioning (and schema versioning in general) is a complex
 topic, involving the need to balance security with developer convenience.
 
 On one end of the spectrum, some use cases require cryptographically bound and 
-immutable contexts, and will need to use a content-addressable URL mechanism,
-such as the in-progress 
-[Hashlink](https://tools.ietf.org/html/draft-sporny-hashlink-03) spec.
+immutable contexts, and will need to enforce immutability at the code level
+(by using a content-addressable URL mechanism such as the in-progress 
+[Hashlink](https://tools.ietf.org/html/draft-sporny-hashlink-03) spec, or by
+forcing the resolution of a given URL to a known hard-coded embedded context).
```
TallTed: I think this should allow for as-yet-unknown code-level options, so `for example` should be added to start or end of the parenthetical.

dmitrizagidulin: Thanks, will update.

OR13:
As I noted on https://github.com/w3c-ccg/did-spec/pull/270

Changes to the context, may cause broken links such as https://web-payments.org/vocabs/security#publicKeyJwk

It would be awesome to add a sentence or two about whether that matters or not, if it does matter, contributors should be guided to submit corrections for:

"sec": "https://w3id.org/security#",

They can read more about how to accomplish that here: https://github.com/perma-id/w3id.org

brentzundel:

msporny:
Migrated to https://github.com/w3c/did-core/pull/91. Closing.

